PR #2, opened 2020-05-30
Zach/dev

zlshames:


src/main/server/helpers/actions.ts
```diff
@@ -0,0 +1,118 @@
...
+        chatGuid: string,
+        message: string,
+        attachmentName?: string,
+        attachment?: Uint8Array
+    ): Promise<Message> => {
+        if (!chatGuid.startsWith("iMessage"))
+            throw new Error("Invalid chat GUID!");
+
+        // Create the base command to execute
+        let baseCmd = `osascript "${this.fs.scriptDir}/sendMessage.scpt" "${chatGuid}" "${message}"`;
```
jarvis-blujay: This builds a shell command by interpolating user-controlled values (`chatGuid`, `message`, and later `attachmentName`) and executes it via `exec`. A crafted quote or shell metacharacter can break argument boundaries and execute arbitrary commands. Use `spawn`/`execFile` with an argument array instead of string concatenation.

src/main/server/fileSystem/index.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Saves an attachment
+     *
+     * @param name Name for the attachment
+     * @param buffer The attachment bytes (buffer)
+     */
+    saveAttachment(name: string, buffer: Uint8Array): void {
+        fs.writeFileSync(path.join(this.attachmentsDir, name), buffer);
```
jarvis-blujay: `name` is passed straight into `path.join(this.attachmentsDir, name)` from socket input. Values like `../...` can escape `attachmentsDir` and write arbitrary files on disk. Please normalize and enforce that the resolved path stays under `attachmentsDir` (or reject any path separators in `name`).

src/main/server/index.ts
```diff
@@ -0,0 +1,368 @@
...
+    private startChatListener() {
+        // Create a listener to listen for new/updated messages
+        const newMsgListener = new MessageListener(this.iMessageRepo, DEFAULT_POLL_FREQUENCY_MS);
+        const updatedMsgListener = new MessageUpdateListener(this.iMessageRepo, DEFAULT_POLL_FREQUENCY_MS);
+
+        newMsgListener.on("new-entry", async (item: Message) => {
+            // ATTENTION: If "from" is null, it means you sent the message from a group chat
+            // Check the isFromMe key prior to checking the "from" key
+            const from = (item.isFromMe) ? "yourself" : item.handle?.id
+            const text = (item.cacheHasAttachments) ? (
```
jarvis-blujay: `item.text` can be null for attachment-only messages. Calling `.slice(...)` before the fallback causes a runtime throw in the listener path. Guard with optional chaining (e.g. `item.text?.slice(...)`) before applying the `<No Text>` fallback.

src/main/server/services/socket/index.ts
```diff
@@ -0,0 +1,394 @@
...
+                console.log("Closing client connection. Authentication failed.");
+            }
+
+            /**
+             * Error handling middleware for all Socket.IO requests.
+             * If there are any errors in a socket event, they will be handled here.
+             *
+             * A console message will be printed, and a socket error will be emitted
+             */
+            socket.use((packet, next) => {
```
jarvis-blujay: This middleware only catches synchronous throws around `next()`. Most handlers below are async and perform DB/fs/network calls, so rejected promises will bypass this block and can become unhandled rejections. Wrap each async handler with a shared `try/catch` helper (or `Promise.resolve(handler).catch(...)`) to guarantee error responses.

zlshames:
while these are valid, why comment and run this on a branch merge from 6 years ago?

PR #216, opened 2021-09-01
Add reply to guid to data sent to clients

tneotia:


zlshames:
Not merging yet until we can investigate how to not use columns with macos versions that don’t support a specific column

zlshames:
I've implemented this conditionally based on os version

PR #251, opened 2021-11-11
Private API support for multiple users

elliotnash:
This goes along with my helper bundle pr and changes the server to use the same port as the bundle


tneotia (approved):

zlshames:
I will merge this once we have a bundle that has this feature built in

elliotnash:
The bundle already has this in master, it'd be nice to merge this + and make a server build at the same time the next bundle build is out

tneotia:
I haven't built a bundle with this but someone can ping me and I can cook one up whenever we want to make a new release. Probably once the mentions support is finalized.

PR #254, opened 2021-11-18
deserialize attributedBody field

elliotnash:
This pr adds a swift helper which can convert the serialized objective C classes in the attributedBody column to json, so mentions and other text metadata can be read
(needs testing on catalina and lower + ~~a way to distribute the helper binary in releases~~ DONE)


tneotia:
So whats the issue with distributing the helper binary? Can we not include it in the server code and call it that way? I may be oversimplifying, I don't really know much about how the Swift helper works.

elliotnash:
If you mean running the helper process from node js, that's what I'm doing, I just hadn't gotten around to packaging the helper binary in the release binary of the server, but I just switched (commit 
bca281a) it to load the helper binary from the appResources folder the same way the private api bundles are copied

zlshames (changes requested):
Hey, so a few things:

1. We need to remove your changes to the bundle. The bundle stuff should only really be updated by me as to prevent compromising the code in it.
2. We need to remove the compiled SwiftHelper. I'll do the compiling

src/main/server/api/v1/interfaces/messageInterface.ts
```diff
@@ -37,6 +37,7 @@ export class MessageInterface {
     static async sendMessageSync(
         chatGuid: string,
         message: string,
+        attributedBody: Record<string, any>,
```
zlshames, line 40: This needs to be optional. This interface handler is for both the private API and AppleScript. So we need to make it optional since it's not something that the AppleScript can implement

src/main/server/databases/transformers/AttributedBodyTransformer.ts
```diff
@@ -0,0 +1,7 @@
+import { Server } from "@server/index";
+import { ValueTransformer } from "typeorm";
+
+export const AttributedBodyTransformer: ValueTransformer = {
+    from: dbValue => Server().swiftHelper.deserializeAttributedBody(dbValue),
```
zlshames, line 5: If the `dbValue` is null, will this error out?

zlshames (changes requested):

src/main/server/api/v1/interfaces/messageInterface.ts
elliotnash: That was a typing error, sendMessageSync still works on a null value, but should be resolved now

src/main/server/databases/transformers/AttributedBodyTransformer.ts
elliotnash: It won't error out, however it will still send it to the swift helper (which probably isn't ideal), I'll add a null check in a bit

elliotnash:
Swift helper is removed :)
The reason I don't want to integrate this into the private API bundle is that this is possible without disabling SIP
For the bundle stuff, I didn't change the helper binary, I just renamed the parent folder to macos11+ since it also supports MacOS 12

src/main/server/services/helperProcess/index.ts
```diff
@@ -34,8 +34,7 @@ export class BlueBubblesHelperService {
         const pApiEnabled = Server().repo.getConfig("enable_private_api") as boolean;
         if (!pApiEnabled) return;
 
-        // eslint-disable-next-line no-nested-ternary
-        const macVer = isMinMonteray ? "macos12" : isMinBigSur ? "macos11" : "macos10";
+        const macVer = isMinBigSur ? "macos11+" : "macos10";
```
zlshames, line 37: I'd like to keep this just as macos11 (without the +)

zlshames (commented):
Can you undo your macos11+ changes

src/main/server/services/helperProcess/index.ts
elliotnash: Would you like me to just rename the bundle back to macos11, or should I also revert the change to allow the macos11 bundle to be installed on Monterey?

zlshames:
@elliotnash yes please revert the bundle changes as well

elliotnash:
All the bundle changes are reverted + up to date with development branch

zlshames:
> All the bundle changes are reverted + up to date with development branch

Thank you. Sorry, I've been super busy for this holiday season... will look to get back on track at some point soonish

zlshames (approved):

swiftHelper/.swiftpm/xcode/xcshareddata/xcschemes/swiftHelper.xcscheme
```diff
@@ -0,0 +1,84 @@
...
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "swiftHelper"
+            BuildableName = "swiftHelper"
+            BlueprintName = "swiftHelper"
+            ReferencedContainer = "container:">
+         </BuildableReference>
+      </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "&quot;/Users/elliot/Library/Application Support/Electron/swift-helper.sock&quot;"
```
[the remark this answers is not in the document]
zlshames: Yeah this is outdated now

PR #279, opened 2021-12-23
Feat/add basic ngrok auth

jlesko2:
## The problem
Users are unable to set basic user/pass on their Ngrok tunnels.
## What your code solves
My code takes advantage of the Ngrok feature to enable basic user:pass authentcation on HTTP tunnels.
## How you fixed it
I added an additional switch-toggle (only available when users provide their own Ngrok auth key) that when enabled, prompts for user/pass with 2 text boxes. Once the user clicks the button `Enable Ngrok Auth` it adds  the additional user:pass to the ngrok connection `opts` and reloads the connection

## Notes:

1. I'm having difficulty sizing the svg next to the ngrok password textbox
2. This change is only half of the solution, the client side must be modified as well!


zlshames:
> I'm having difficulty sizing the svg next to the ngrok password textbox

Do you have a screenshot of what it looks like in the current state?

> This change is only half of the solution, the client side must be modified as well!

I'm glad you recognize this. I will leave this PR open until there is a stable release containing code on the client side to support it. We probably need to do a few things client side:

* Make the toggle for basic auth show on the setup screen & add required user/pass text fields
* Add the toggle for basic auth to the server settings, along with required user/pass text fields (when enabled)
* API detection to "figure out" if basic auth is enabled, but not enabled on the client side. Show an error or warning if that's the case.

zlshames:
Hey, so i did a complete server rewrite. If you'd still like this implemented, can you update your PR? I think you'll find the new UI to be much easier to work with too :)

jlesko2:
> 

Thanks for the FYI! I would be happy to. Sorry this is taking so long- I'm struggling to  on the client side. I can't seem to get the dart socket_io client to authenticate against a tunnel with basic auth. I created a small dart test file with the same socket_io library and it can authenticate without issue. I'll raise this in the discord when I am in a good spot again.

jlesko2:
> Hey, so i did a complete server rewrite. If you'd still like this implemented, can you update your PR? I think you'll find the new UI to be much easier to work with too :). The code is in the development branch btw

Thanks for the FYI @zlshames ! I would be happy to. Sorry this is taking so long- I'm struggling to  on the client side. I can't seem to get the dart socket_io client to authenticate against a tunnel with basic auth. I created a small dart test file with the same socket_io library and it can authenticate without issue. I'll raise this in the discord when I am in a good spot again.

PR #289, opened 2022-01-23
Decrease Database Receive poll to 100miliseconds

mrsylerpowers:
# The Problem

Message responses can take up to 1 second before being sent to clients

# What your code solves

My code decreases the time between message queries and allows for a much faster-receiving rate of messages to clients

# How you fixed it
Decreased the default poll time for all message listeners

# Notes:
This may affect the outgoing message listener in an unpredictable way, there are multiple hardcoded artificial delays bigger or the same as 200ms inside of the outgoing lister.

### DIFF
--- a/src/main/server/constants.ts
+++ b/src/main/server/constants.ts
@@ -1,1 +1,1 @@
-export const DEFAULT_POLL_FREQUENCY_MS = 1000;
+export const DEFAULT_POLL_FREQUENCY_MS = 100;


zlshames:
I avoided checking the DB too quickly because I didn't want to flood the database with read actions. Do you know if this may effect the PCs performance or native iMessage performance?

mrsylerpowers:
@zlshames I am not seeing a significant performance hit. I created a new change listener as a test. This change listener measured the processing time for each query taking normally 2ms for null result queries and up to 5ms when finding and processing a new message. Meaning the listener is still waiting lots of time and not completely saturating with reads.

zlshames:
See my previous comment on your other PR. The only difference is i made it so 200ms was the minimum. you can't go lower

PR #296, opened 2022-02-07
Remove message match race condition POC

mrsylerpowers:
MessageMatcheEventEmits should be done on completion of its attached MessagePromise. Artificial delays can be removed from the OutgoingMessageListener as long as this line `Server().messageManager.promises[idx].resolve(entry)` resolves after its message match event has been emitted to clients.


zlshames:
Is this still relevant or should you close it?

mrsylerpowers:
@zlshames Set it up to auto close

PR #298, opened 2022-02-07
Fix Private Messaging Updates

mrsylerpowers:
There is a bug where private messages don't update their status. This is because private messages send so fast that a new race condition existed between the outgoing message listener and the sending of private messages. If private messages were sent and delivered faster than the speed that the outgoing message listener refreshes its database the new outgoing message listener would cache the 'new message'. This message would then block updates to themselves and others because of the cache return statements.  A race to be the first in the cache and new messages almost always win

Notes:  There is still a possibility for a race condition where an updated message listener triggers on a new-message


zlshames:
Is this still relevant or should you close it?

PR #310, opened 2022-03-05
Add ability for message update from private api signal

mrsylerpowers:


zlshames (commented):
You should make it so it passes a message guid back to the server so that we can fetch the message and send the update event. Rather than cancelling the check and re-checking. Especially since the check happens every 1 second about

PR #371, opened 2022-06-09
Incoming facetime call detection

SushiPython:
Polls for incoming facetime calls and then sends an FCM
(Requires client modifications for it to actually do anything)

The problem
-No facetime detection
What your code solves
-Adds facetime detection
How you fixed it
-Polling the mac notifications for incoming facetime calls


zlshames:
I'm going to look at this more in depth as we update the client to support the new message type. Once it's supported, or on track to be supported, I can merge this. I didn't forget about it :)

PR #416, opened 2022-09-02
v1.3.0

zlshames:


zlshames:
# What's New?

## Fixes & Improvements

* Improves image & video sending on macOS Monterey & newer
    - We discovered a workaround for sending attachments the same way we do on older macOS versions, which should mean the reliability of sending attachments is fairly high.
    - This also means Accessibility is no longer required for macOS Monterey+.
* Adds ability to manage your BlueBubbles attachment cache on the server (Settings).
* Fixes issue where BlueBubbles would be detected as being from an "unidentified" developer.

PR #466, opened 2022-12-12
Automate the Firebase setup

probablypablito:
Uses Firestore instead of Realtime DB

### DIFF
--- a/temp/firebase-setup.ts
+++ b/temp/firebase-setup.ts
@@ -1,0 +1,266 @@
+const express = require("express")
+const app = express()
+
+// Express stuff is safe to remove, used it for testing
+
+const {google} = require("googleapis")
+
+import axios from "axios"
+import { googleClientSecret } from "./keys/keys";
+import { googleClientID } from "./keys/keys";
+
+const googleOauthRedirectURL: string = "http://localhost:3000/api/login/google"
+
+const oauth2Client = new google.auth.OAuth2(
+     googleClientID,
+     googleClientSecret,
+     googleOauthRedirectURL
+);
+
+
+
+
+let projectName: string, token: string // Define in a global scope
+
+app.get("/", async (req: any, res: any) => res.send(`<a href=${await generateAuthUrl()}>Sign in with Google</a>`))
+app.get("/api/login/google", async (req: any, res: any) => {
+
+    const code = req.query.code
+    const tokensJSON = await oauth2Client.getToken(code)
+    token = tokensJSON.tokens.access_token
+
+    console.log("token: " + token) // For testing with reqbin, should be removed
+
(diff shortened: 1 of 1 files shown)


zlshames:
@probablypablito this looks great! I'll need to convert it to a service class. One thing, did you implement a way to get the Client JSON? I see you implemented a way to get the services JSON, but how about the client JSON

zlshames:
Also, any chance you can implement the ability to check if a project with the selected name exists, and if it does, just fetch the JSONs for the existing one? Basically, I want the following flow when setting up the server:

1. Open the server for the first time
2. Click button to sign in with google
3. When the callback fires, check to see if a project exists, and if so, download the JSONs and move on with the setup.
4. If the project doesn't exist, forward to a screen that shows the progress of the app setting up. Then downloads the JSON and moves on to the next step when done.

You dont have to worry about all that. But being able to check if a project exists already would help that workflow

probablypablito:
I'm not sure what you mean by the Client JSON, I only did the ones that you have to drag and drop into BB currently.

To check if a project exists you can use [this](https://cloud.google.com/resource-manager/reference/rest/v1/projects/get) GET request. Code 200 if it exists, 403 if it doesn't. If it does running the `getGoogleServicesJson` and `getPrivateKey` will work

temp/firebase-setup.ts
```diff
@@ -0,0 +1,266 @@
...
+app.get("/", async (req: any, res: any) => res.send(`<a href=${await generateAuthUrl()}>Sign in with Google</a>`))
+app.get("/api/login/google", async (req: any, res: any) => {
+
+    const code = req.query.code
+    const tokensJSON = await oauth2Client.getToken(code)
+    token = tokensJSON.tokens.access_token
+
+    console.log("token: " + token) // For testing with reqbin, should be removed
+
+    projectName = "bluebubblestest43" //  Must be 6 to 30 lowercase letters, digits, or hyphens. It must start with a letter. Trailing hyphens are prohibited.
```
mrsylerpowers, line 34: Used as globals injected into functions?

temp/firebase-setup.ts
```diff
@@ -0,0 +1,266 @@
...
+
+
+let projectName: string, token: string // Define in a global scope
+
+app.get("/", async (req: any, res: any) => res.send(`<a href=${await generateAuthUrl()}>Sign in with Google</a>`))
+app.get("/api/login/google", async (req: any, res: any) => {
+
+    const code = req.query.code
+    const tokensJSON = await oauth2Client.getToken(code)
+    token = tokensJSON.tokens.access_token
```
mrsylerpowers, line 30: Used as globals injected into functions?

zlshames:
Closing as this is now implemented natively in my branch `zach/google-login`

PR #474, opened 2023-01-19
Handle fix

mrsylerpowers:
Fix not buildable on on macos 10.15sdk


zlshames:
Just confirmed. The node Mac permission downgrade shouldn't affect compatibility with the APIs we use. It was mostly chore downgrades

PR #485, opened 2023-02-13
fix(server): quit FindMy app 2min after refresh

keegan-lillo:
## The problem
- The FindMy service stops updating if the app is hidden for a while

Context: https://discord.com/channels/640229510662455326/725057224908472381/1073937589138096179

## What your code solves
- Adds a 2 minute timeout to quit the FindMy app after performing a refresh
- Updates the readme to have the correct dev start script
- Cleaned up `packages/server/src/server/api/v1/apple/scripts.ts` a little

## How you fixed it
- Updated the `FindMyService.refresh()` method to start a 2 minute timeout to quit the FindMy app

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This is the back-end server for the BlueBubbles App. It allows you to forward yo
 3. Install the server dependencies
     - `yarn`
 4. Run the dev server (this will start both the renderer and server)
-    - `yarn run start-dev`
+    - `yarn start`
 
 ## Structure / Directory Map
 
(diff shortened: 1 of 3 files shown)


zlshames:
@keegan-lillo Instead of quitting FMF after hitting the refresh endpoint, wouldn't it make more sense to just quit it before re-opening it? And we can track the "last refresh" time to make sure it doesn't get "quit" more than once within 2 minutes?

I bring it up because what if the user uses the mac and has FMF open already, they wouldnt want it to be quit randomly  and not reopen until refresh is called again, right?

keegan-lillo:
Ah true. Didn't consider that. Okay I'll update.

keegan-lillo:
@zlshames okay updated! Just an FYI: This will add an extra 3 seconds to the request time while we wait for the app to restart.

zlshames:
LGTM

PR #592, opened 2023-11-21
V2 API with Tokens!

juliettebee:
Hiya! So I noticed BlueBubbles doesn't use tokens and has the interesting design choice of using url params instead of headers so I created a version 2 of the API with user created tokens! **This is very much a proof of concept, it has not been thoroughly tested, it was hacked together in 4 hours**. This is the first time I've written in Typescript so lemme know if something's wrong! 

## Changes:
* Token panel in settings
* V2 API requiring a `Token` header with the value of `[token name] [token password]`
* Two new API routes
     - `token/register` POST `{"name": "token name", "password": "token password"}`
     - `token/delete` POST `{"name": "token name"}`


juliettebee:
Added route: `token/refresh` POST `{"name": "token name", "password": "token password", expireAt: 0}`
Where expireAt is how much longer to add to the expiration in Unix time:)

zlshames:
Sorry I havent looked into this yet. I'm waiting for the right time to properly implement it

packages/ui/src/app/constants.ts
zlshames, line 1: Definitely a cool idea to add an event for when new tokens are registered for monitoring purposes. Good idea.

zlshames (changes requested):
Thanks for submitting this of course! I have some changes that I think would be good to make. It may complicate some of the logic in some cases, but using JWTs and being able to refresh a token is typical in many Bearer auth flows

PR #636, opened 2024-04-05
Native Apple Silicon/arm-64 support

### LINKED ISSUE #383: Native apple silicon support
Could we get a server app that supports apple silicon natively or is a universal binary? I am noticing that intel apps use a large amount of ram (around 500 MB) at idle, while native apps consume closer to 200mB at idle. Many thanks for considering my request.

CraftyCorvid:
Closes #383

### DIFF
--- a/packages/server/scripts/electron-builder-config.js
+++ b/packages/server/scripts/electron-builder-config.js
@@ -38,7 +38,7 @@ module.exports = {
                 "target": "default",
                 "arch": [
                     'x64',
-                    // 'arm64'
+                    'arm64'
                 ]
             }
         ],
--- a/packages/server/src/server/managers/cloudflareManager/index.ts
+++ b/packages/server/src/server/managers/cloudflareManager/index.ts
@@ -9,7 +9,9 @@ import { Loggable } from "@server/lib/logging/Loggable";
 export class CloudflareManager extends Loggable {
     tag = "CloudflareManager";
 
-    daemonPath = path.join(FileSystem.resources, "macos", "daemons", "cloudflare", "cloudflared");
+    daemonPath = path.join(
+        FileSystem.resources, "macos", "daemons", "cloudflare", `cloudflared${process.arch === 'arm64' ? '-arm64' : ''}`
+    );
 
     // Use a default (empty) config file so we don't interfere with the default CF install (if any)
     cfgPath = path.join(FileSystem.resources, "macos", "daemons", "cloudflare", "cloudflared-config.yml");


zlshames:
I'm not 100% sure this alone will fix it. I also think I need an arm mac to compile this properly for apple silicon. I've tried compiling on my intel mac, and it works, but crashes on startup due to dependency linking and stuff

craftycorvid:
I've been running a build with these changes for a couple weeks without issues on my M2 MacBook.

You should be able to compile it on your Intel mac, electron has built-in support for cross-compiling. I'd be happy to test a build of yours to make sure it runs correctly.

zlshames:
> I've been running a build with these changes for a couple weeks without issues on my M2 MacBook.
> 
> You should be able to compile it on your Intel mac, electron has built-in support for cross-compiling. I'd be happy to test a build of yours to make sure it runs correctly.

If you join our Discord and DM me, I'm down to send you an arm64 build built by me.

duronald:
Are there any updates on this? I'm happy to test too.

zlshames:
I've implemented this into the development branch. There is no cloudflared Apple Silicon binary though, so that will still run using rosetta. Same with ngrok and zrok (i'm assuming).

craftycorvid:
There's a cloudflared arm64 binary in this PR that I got from homebrew: https://formulae.brew.sh/formula/cloudflared

zrok also has one there: https://formulae.brew.sh/formula/zrok#default

ngrok seem to be stuck on Intel.

craftycorvid:
Actually ngrok has an apple silicon build on their downloads page: https://ngrok.com/download

craftycorvid:
> I've been using: https://github.com/cloudflare/cloudflared/releases
> 
> How come homebrew has an apple silicon build, but the github doesn't 🤔
> 
> I also see that the builds say, `arm64-sonoma` or `arm64-ventura`. Does that mean those specifically run on those macos versions? or what? what if i run the ventura one on big sur?

It's unclear why they don't have a build on their release page: https://github.com/cloudflare/cloudflared/issues/534

Homebrew builds against each version of MacOS in case built-in dependencies changed. The build against the oldest version of MacOS you want to support is probably the correct one.

zlshames:
> > I've been using: https://github.com/cloudflare/cloudflared/releases
> > How come homebrew has an apple silicon build, but the github doesn't 🤔
> > I also see that the builds say, `arm64-sonoma` or `arm64-ventura`. Does that mean those specifically run on those macos versions? or what? what if i run the ventura one on big sur?
> 
> It's unclear why they don't have a build on their release page: [cloudflare/cloudflared#534](https://github.com/cloudflare/cloudflared/issues/534)
> 
> Homebrew builds against each version of MacOS in case built-in dependencies changed. The build against the oldest version of MacOS you want to support is probably the correct one.

right, but the oldest one is ventura, but this binary needs to support all the way down to high sierra. I know the official builds do, but im not 100% sure of these builds

craftycorvid:
Big Sur is the furthest back that arm64 support could possibly go, but I get your point. Homebrew seems to drop older versions of MacOS pretty quickly.

zlshames:
> Big Sur is the furthest back that arm64 support could possibly go, but I get your point. Homebrew seems to drop older versions of MacOS pretty quickly.

i was more meaning for the x86 builds. However, I guess I can get the x86 build from the github, but then use the homebrew build for the arm64 build. That's probably a good solution

zlshames:
@craftycorvid do you know how i can download the raw binary from homebrew's formulae website? when i try using the link in the JSON it fails with a 401

craftycorvid:
I had to download it using the package manager. They're doing some sort of authentication on their links that the package manager handles.

zlshames:
@craftycorvid i ended up following this guide: https://github.com/orgs/Homebrew/discussions/4335#discussioncomment-5353698

A bit annoying... but not too bad I guess

PR #680, opened 2024-08-11
Update arm64 cloudflared build

pedrosax:
Update arm64 (Apple Silicon) cloudflared binary to the official arm64 build from Cloudflare.


zlshames:
So, unfortunately, I've already implemented this in the server on my development branch:

https://github.com/BlueBubblesApp/bluebubbles-server/tree/development/packages/server/appResources/macos/daemons/cloudflare

and

https://github.com/BlueBubblesApp/bluebubbles-server/blob/development/packages/server/src/server/managers/cloudflareManager/index.ts#L11-L12

PR #690, opened 2024-09-02
Make sure the machine never sleeps...ever

cameronaaron:
I noticed even with amphetamine my machine would randomly decide to sleep and it would cause the server to break so I took a stab at making it so even if there was a nuclear war this thing would be a roach and not die

### DIFF
--- a/packages/server/src/server/services/caffeinateService/index.ts
+++ b/packages/server/src/server/services/caffeinateService/index.ts
@@ -1,6 +1,7 @@
 import * as process from "process";
 import { spawn, ChildProcessWithoutNullStreams } from "child_process";
 import { Loggable } from "@server/lib/logging/Loggable";
+import { setInterval } from "timers";
 
 /**
  * Service that spawns the caffeinate process so that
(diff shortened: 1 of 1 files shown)


mehapps:
I think you could safely change the watchdog interval to something like 30 seconds or even a minute? I'm sure there's probably a good reason you did 5 seconds but AFAIK the lowest sleep interval is a minute of inactivity?

cameronaaron:
I might change it to 30 largely the thing I really want to prevent is it detecting the laptop is closed if it crashes during that time

mehapps:
I see, would it maybe be safer to use pmset to disable sleep entirely? That way caffeinate doesn't have to be a running subprocess, and it could just be reverted if u disable the setting in bluebubbles

cameronaaron:
I was thinking about that actually...That honestly might be the way to go...instead of caffinate we just kill sleep then revert when the box is unchecked I think @zlshames would be more in favor of that as well

mehapps:
FWIW, these commands should do it (from OpenCore's Fixing Sleep guide + my own experience)
```
sudo pmset -a sleep 0
sudo pmset autopoweroff 0
sudo pmset powernap 0
sudo pmset standby 0
sudo pmset tcpkeepalive 0
sudo pmset proximitywake 0
sudo pmset -a disablesleep 1
```

zlshames (commented):
So I don't love the idea of preventing the display from sleeping. Some people might want to let their display turn off, cuz everything else will remain working. Also, what's the point of using the Run Indefinitely flag? Doesn't it already do that by default? Or is the default a lower limit?

zlshames:
> FWIW, these commands should do it (from OpenCore's Fixing Sleep guide + my own experience)
> 
> ```
> sudo pmset -a sleep 0
> sudo pmset autopoweroff 0
> sudo pmset powernap 0
> sudo pmset standby 0
> sudo pmset tcpkeepalive 0
> sudo pmset proximitywake 0
> sudo pmset -a disablesleep 1
> ```

The server can't run sudo commands

cameronaaron:
It's been a min since I looked at this but I think I was hoping to essentially cut out the need for amphetamine especially on a laptop but since then focus on tricking it into thinking it's in clam shell.

mehapps:
> > FWIW, these commands should do it (from OpenCore's Fixing Sleep guide + my own experience)
> > ```
> > sudo pmset -a sleep 0
> > sudo pmset autopoweroff 0
> > sudo pmset powernap 0
> > sudo pmset standby 0
> > sudo pmset tcpkeepalive 0
> > sudo pmset proximitywake 0
> > sudo pmset -a disablesleep 1
> > ```
> 
> The server can't run sudo commands

Ah I see, is there any way to escalate just when the box is clicked? Like when the box is clicked ask for sudo password and then run those commands? Those are just one time things, so you would only need the password when it's checked or unchecked.

zlshames:
> > > FWIW, these commands should do it (from OpenCore's Fixing Sleep guide + my own experience)
> > > ```
> > > sudo pmset -a sleep 0
> > > sudo pmset autopoweroff 0
> > > sudo pmset powernap 0
> > > sudo pmset standby 0
> > > sudo pmset tcpkeepalive 0
> > > sudo pmset proximitywake 0
> > > sudo pmset -a disablesleep 1
> > > ```
> > 
> > 
> > The server can't run sudo commands
> 
> Ah I see, is there any way to escalate just when the box is clicked? Like when the box is clicked ask for sudo password and then run those commands? Those are just one time things, so you would only need the password when it's checked or unchecked.

I'd need to check. Unsure if it hinges on the app being notarized or just signed. Research required...

mehapps:
Sounds good, afaik it doesn't (from my experience from using python with the elevate module)

cameronaaron:
closing will be superseded by https://github.com/BlueBubblesApp/bluebubbles-server/pull/706

PR #696, opened 2024-09-21
fix notification issue in server

cameronaaron:


cameronaaron:
@zlshames Any thoughts? I want to also do a refactor of the full app because there is a lot of lib stuff that needs to be done because things are being done manually where there are libs but I wanna know your opinion first on this

zlshames:
> @zlshames Any thoughts? I want to also do a refactor of the full app because there is a lot of lib stuff that needs to be done because things are being done manually where there are libs but I wanna know your opinion first on this

I like what I see. I just haven't had a chance to really pull it and try it. I'll get around to it, I just have other priorities right now. 

I'm not sure what you're talking about in terms of the lib stuff, but if you're talking about the app/client, I'm already in the middle of a massive refactor of the codebase in effort to make it better, more efficient and less janky

PR #736, opened 2025-05-01
Update Zrok to 1.0.2

### LINKED ISSUE #731: upgrade zrok binary
errors are being thrown

#### Issue Metadata

- User: `@zach.shames#0`
- Context: https://discord.com/channels/640229510662455326/782690923640782878/1351529927811530784

Scr0nch:
# Purpose
Resolves #731.

# Method
1. Updates the API URL to 1.0
2. Updates the included executable daemon to 1.0.2, which is the latest version at the time of writing
3. Updates documentation about executable daemon's version and how to verify them
4. Update the way the JSON output of the `zrok overview` command's output is parsed: the `token` property was renamed to `shareToken`

# Testing
After building, I configured the server from scratch multiple times and tried the following Zrok configurations:

- Reserved share with no set subdomain
- Reserved share with set subdomain
- Ephemeral share (with no set subdomain)
- Disabling and re-enabling environment
- Register new account button
- Changing reserved share subdomain

The following situations are not handled ideally, but I don't know if they are related to this patch/regressions:
- Invalid Zrok token: configuration is saved but appears to continue to use the last valid token internally even after a server restart, no alerts displayed
- Duplicate share name: configuration is saved successfully but then fails to create the share and errors are displayed in the alerts every automatic retry until the name is changed. A `409 CONFLICT` status code is returned.
- Disabling internet: although the string `"dial tcp: lookup api-v1.zrok.io"` does appear in the alerts, it is logged from classes other than `ZrokManager`; `ZrokManager` has a special case in `handleData` to make the alert message nicer, but this case isn't hit. For example:
  ```
  [ZrokService] Failed to restart Zrok! Error: Error [ERR_UNHANDLED_ERROR]: Unhandled error. ... dial tcp: lookup api-v1.zrok.io ...
  ```
   Other errors are also displayed before the server recovers once internet access is restored:
  ```
  [ZrokManager] Failed to run Zrok daemon! [object Object]
  ```

I've submitted this pull request regardless since these issues do not prevent Zrok integration from working properly and the May 19th deadline to upgrade to 1.0 is approaching quickly.


Scr0nch:
Hey @zlshames, @jjoelj, and @tneotia, I'm pinging you because Zrok will drop support for 0.4 in three days, meaning this must be merged very soon for Zrok integration to stay functional. Is it possible to get this in before then?

I'm fine with using my fork, but others who don't have it will have to switch to a different proxy service after they realize they can no longer connect to their server.

zlshames (approved):

zlshames:
Thank you for all the details & contribution! I'll be merging it, but will immediately upgrade the zrok binaries to 1.0.4 (latest)

Scr0nch:
Thank you! Ideally we would use git large file storage to store these binaries to reduce the size of the repository, but that might require installing git-lfs in the build pipeline and, for completeness, means rewriting the repository history to remove existing binary files using their migration tool, so I didn't bother with this pr.

PR #764, opened 2026-01-14
Add LaunchAgents directory check and mkdir

elsiehupp:
As of macOS Tahoe, the directory `~/Library/LaunchAgents` does not exist unless otherwise created.

Therefore, if the user is setting up BlueBubbles on a brand new system running macOS Tahoe, `createLaunchAgent()` will fail silently if the directory `~/Library/LaunchAgents` has not been manually created.

This Pull Request \*I think\* should fix that. (I'm not a Node developer, so this is my best guess for how to fix this issue.)

### DIFF
--- a/packages/server/src/server/fileSystem/index.ts
+++ b/packages/server/src/server/fileSystem/index.ts
@@ -746,8 +746,14 @@ export class FileSystem {
     </dict>
 </plist>`;
 
+        // As of macOS Tahoe, the directory ~/Library/LaunchAgents does not exist unless otherwise created
+        const launchAgentsPath = path.join(userHomeDir(), "Library", "LaunchAgents");
+        if (!fs.existsSync(launchAgentsPath)) {
+            fs.mkdirSync(launchAgentsPath);
+        }
+
         const plistName = "com.bluebubbles.server";
-        const filePath = path.join(userHomeDir(), "Library", "LaunchAgents", `${plistName}.plist`);
+        const filePath = path.join(launchAgentsPath, `${plistName}.plist`);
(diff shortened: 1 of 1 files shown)


zlshames (approved):
Looks good, simple change. Good call out about Tahoe. Like you've implemented, probably doesn't need to be macOS-version-controlled since it's a user-directory anyays.

elsiehupp:
@zlshames FYI it seems like macOS Tahoe has implemented stricter controls around Launch Agents more generally. Like, among other things you can't use unsigned executables as Launch Agents, [which breaks Beeper Bridge Manager because its prebuilt executables are unsigned](https://github.com/beeper/bridge-manager/issues/69).

If you wanted to poke around with Beeper Bridge Manager to see if you could help them with macOS packaging and code signing, that would be great. I've described the basic outline of a proposed solution in [my Issue there](https://github.com/beeper/bridge-manager/issues/69), but it's outside my general realm of experience.

zlshames:
> @zlshames FYI it seems like macOS Tahoe has implemented stricter controls around Launch Agents more generally. Like, among other things you can't use unsigned executables as Launch Agents, [which breaks Beeper Bridge Manager because its prebuilt executables are unsigned](https://github.com/beeper/bridge-manager/issues/69).
> 
> If you wanted to poke around with Beeper Bridge Manager to see if you could help them with macOS packaging and code signing, that would be great. I've described the basic outline of a proposed solution in [my Issue there](https://github.com/beeper/bridge-manager/issues/69), but it's outside my general realm of experience.

Thanks for the heads-up... Great

PR #766, opened 2026-01-22
Support macOS 15+ text formatting

Nicell:
Adds support for basic text formatting (bold, italic, underline, strikethrough)

Pairs with https://github.com/BlueBubblesApp/bluebubbles-helper/pull/50

This script:

```js
const body = {
  chatGuid,
  message: "Bold Italic Underline Strike",
  method: "private-api",
  textFormatting: [
    { start: 0, length: 4, styles: ["bold"] },
    { start: 5, length: 6, styles: ["italic"] },
    { start: 12, length: 9, styles: ["underline"] },
    { start: 22, length: 6, styles: ["strikethrough"] }
  ]
};

fetch(`${baseUrl}/api/v1/message/text?password=${encodeURIComponent(password)}`, {
  method: "POST",
  headers: { "Content-Type": "application/json" },
  body: JSON.stringify(body)
})
```

Resulted in: 
image

### DIFF
--- a/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
+++ b/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
@@ -14,6 +14,7 @@ export class PrivateApiMessage extends PrivateApiAction {
         chatGuid: string,
         message: string,
         attributedBody: Record<string, any> = null,
+        textFormatting: Record<string, any> = null,
         subject: string = null,
         effectId: string = null,
         selectedMessageGuid: string = null,
@@ -29,6 +30,7 @@ export class PrivateApiMessage extends PrivateApiAction {
             subject,
             message,
             attributedBody,
+            textFormatting,
             effectId,
             selectedMessageGuid,
             partIndex
--- a/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
+++ b/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
@@ -4,7 +4,7 @@ import {
     TransactionType
 } from "@server/managers/transactionManager/transactionPromise";
 import { PrivateApiAction } from ".";
-import type { ValidTapback, ValidRemoveTapback } from "@server/types";
+import type { TextFormatting, ValidTapback, ValidRemoveTapback } from "@server/types";
 import { isMinCatalina, isMinMonterey } from "@server/env";
 
 export class PrivateApiMessage extends PrivateApiAction {
@@ -14,7 +14,7 @@ export class PrivateApiMessage extends PrivateApiAction {
         chatGuid: string,
         message: string,
         attributedBody: Record<string, any> = null,
-        textFormatting: Record<string, any> = null,
+        textFormatting: TextFormatting = null,
         subject: string = null,
         effectId: string = null,
         selectedMessageGuid: string = null,
--- a/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
+++ b/packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
@@ -4,7 +4,8 @@ import {
     TransactionType
 } from "@server/managers/transactionManager/transactionPromise";
 import { PrivateApiAction } from ".";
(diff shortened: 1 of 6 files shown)


packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
```diff
@@ -14,6 +14,7 @@ export class PrivateApiMessage extends PrivateApiAction {
         chatGuid: string,
         message: string,
         attributedBody: Record<string, any> = null,
+        textFormatting: Record<string, any> = null,
```
zlshames: Should this be using the TextFormatting type? The `Record` type would imply the param is a dictionary, right? It's meant to be a list, to support multiple stylings over different ranges, right?

zlshames (commented):
Overall, this looks great! Thanks for doing all of the validation, min MacOS version, and what not, not only the endpoint & calls to the PAPI. Check out my comment, but that's the only comment I have. Obviously, this will need to be paired with a release for the PAPI bundle.

packages/server/src/server/api/privateApi/apis/PrivateApiMessage.ts
Nicell: fixed! thanks

kate-jiang:
this and #768 are huge for my use case !! great work boys

arimendelow:
@zlshames, looking for an update on this — would be an incredible unlock to have this :)

zlshames:
> @zlshames, looking for an update on this — would be an incredible unlock to have this :)

almost done with my rewrite. I'm creating a plan to add this to the client side once i finish ironing out the rewrite. Don't worry, i haven't forgotten. Just have some loose ends to clean up before getting to it.

omarshahine:
Verified this PR end-to-end on macOS 26.4.1 (Apple Silicon) paired with [bluebubbles-helper#50](https://github.com/BlueBubblesApp/bluebubbles-helper/pull/50). When both are applied, `attributedBody.runs` on the response come back with `__kIMTextBoldAttributeName` / `__kIMTextItalicAttributeName` / `__kIMTextStrikethroughAttributeName` on the requested ranges, and the recipient iPhone renders the styles. Nice work.

One release-coordination note worth flagging before this lands: when this PR merges and a new server release is cut, the bundled helper at `packages/server/appResources/private-api/macos11/BlueBubblesHelper.dylib` (and its `.md5`) also needs to be regenerated from helper#50. The dylib currently in this repo is built from helper `master` and does not contain `+[BlueBubblesHelper applyTextFormatting:toMessage:]`, so a server release shipping just this PR would accept the new field on `/api/v1/message/text`, return 200, and then silently render plain text — the helper drops the unknown payload key.

The symptom is hard to debug from outside because nothing logs. The minimal probe is:

```bash
PW=$(sqlite3 "$HOME/Library/Application Support/bluebubbles-server/config.db" \
  "SELECT value FROM config WHERE name='password';")
curl -s -X POST "http://127.0.0.1:1234/api/v1/message/text?password=$PW" \
  -H 'content-type: application/json' \
  -d '{"chatGuid":"any;-;+15555550123","tempGuid":"smoke","method":"private-api","message":"x bold word y","textFormatting":[{"start":2,"length":9,"styles":["bold"]}]}' \
  | jq '.data.attributedBody[0].runs'
```

— a single run with only `__kIMMessagePartAttributeName: 0` means the helper isn't applying styles; presence of `__kIMTextBoldAttributeName: 1` over the right range means end-to-end is good.

Suggested merge order: helper#50 → cut helper release → re-vendor the bundled dylib in this repo → land this PR.

(One small build-environment FYI in case anyone tries to repro the helper build outside Xcode: macOS Sequoia/Tahoe `Messages.app` is `x86_64 + arm64e`, so the rebuilt dylib must include `arm64e` — an `arm64`-only slice silently fails to load via `DYLD_INSERT_LIBRARIES` and the only surface signal is `helper_connected: false`. Already correct on `master`'s shipped dylib; just worth noting for whoever rebuilds.)

zlshames (approved):

PR #767, opened 2026-01-22
fix: stopTyping endpoint was calling startTyping

Nicell:
## Summary

One-line fix: the `DELETE /chat/:guid/typing` endpoint (`stopTyping`) was calling `ChatInterface.startTyping()` instead of `ChatInterface.stopTyping()`.

## Problem

When clients call the stop typing API, typing indicators would restart instead of stop.

## Fix

```diff
-        await ChatInterface.startTyping(guid);
+        await ChatInterface.stopTyping(guid);
```

## Testing

Tested locally - typing indicators now properly stop when DELETE is called.


allycake:
Can bluebubble on iPhone

PR #780, opened 2026-03-19
feat: per-thread webhook filtering by chat GUID

f1sherman:
## Summary

Allow custom webhooks to be configured for each iMessage conversation.

## Context

First of all, this is an excellent project, thank you for building it!

I've been running a few openclaw agents using BlueBubbles with separate iMessage Guids per agent for the communication channel and it has been working flawlessly. However, because all webhook requests go to the same endpoint(s), it makes agent isolation pretty difficult. This change enables users to configure separate webhooks per chat Guid. I would imagine this opens up additional use cases as well.

image

image

## Test plan

- [x] Create a webhook without selecting any chats - verify it receives all events (existing behavior unchanged)
- [x] Create a webhook scoped to a specific 1:1 chat - verify it only receives events from that chat
- [x] Create a webhook scoped to a group chat - verify filtering works for group events
- [x] Verify chat-filtered webhooks still receive global events (`server-update`, `new-server`)
- [x] Edit an existing webhook to add/remove chat filter - verify changes persist
- [x] Verify the chat dropdown shows readable names (display names for groups, addresses for 1:1)
- [x] Upgrade from a previous version - verify migration adds the column and existing webhooks continue working
- [x] Create a webhook via REST API with `chatGuids` parameter - verify round-trip through GET

### DIFF
--- a/packages/server/src/server/api/http/api/v1/routers/webhookRouter.ts
+++ b/packages/server/src/server/api/http/api/v1/routers/webhookRouter.ts
@@ -13,20 +13,26 @@ export class WebhookRouter {
         const id = (ctx?.request?.query?.id) ? Number.parseInt(ctx?.request?.query?.id as string) : null;
         const webhooks = await Server().repo.getWebhooks({ url, id });
 
-        // Convert the events to a list (from json array)
+        // Convert JSON strings to arrays for the response
         for (const webhook of webhooks) {
             webhook.events = JSON.parse(webhook.events);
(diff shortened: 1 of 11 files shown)


zlshames:
Sorry for the extremely late review... I've been working on the Android app trying to get the rewrite out the door. For the UI's chat filter element, is it a freeform field that you can turn the entires into "chips", or does it let you search by existing address in the database?

Also, if you don't mind, resolving the conflict here. Probably just a lint/formatting issue

PR #788, opened 2026-04-04
feat: improve contacts permissions UI and add TCC grant

ThePieMonster:
- Add askForContactsAccess() call when requesting contact permission, opens System Settings to the Contacts pane as fallback
- Add grantContactsViaTCC() helper that uses osascript with admin privileges to grant contacts access via the system TCC database, auto-detecting dev (com.github.Electron) vs prod bundle ID
- Add Contacts to checkPermissions() with dev/prod aware solution text
- Redesign Contacts page layout into three distinct sections: Manage, macOS Contacts permissions, and Google Contacts with individual status indicators for each

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,15 @@ This is the back-end server for the BlueBubbles App. It allows you to forward yo
 4. Run the dev server (this will start both the renderer and server)
     - `npm run start`
 
+### Development Ports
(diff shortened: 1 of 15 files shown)


tneotia:
Any reason why this was closed?

PR #805, opened 2026-05-31
fix: reduce message delivery latency

123qwe55aa:
## Summary
Reduce iMessage delivery latency by addressing idle connection钝化 and XPC断连 timing issues.

## Changes
- **5s heartbeat poll** in Private API helper instead of 60s backfill cycle
- **Reconnect sweep** fires immediately on XPC reconnection to backfill missed messages
- **30s keepalive** keeps connections hot, preventing Apple TCC timeout
- Fixed race condition where `IMessageListener.poll()` waited for async message handlers

## Why
Without keepalive, long idle connections become cold — new messages wait for next poll cycle (up to 60s). With XPC断连 being unavoidable on macOS, the reconnect sweep ensures missed messages are recovered promptly.

## Testing
- Jest tests pass
- TypeScript compiles clean
- macOS arm64 build succeeds
- Local testing confirms sub-5s delivery after idle period

### DIFF
--- a/packages/server/scripts/electron-builder-config.js
+++ b/packages/server/scripts/electron-builder-config.js
@@ -2,7 +2,7 @@
 // Making them relative to the scripts folder will break the commands
 module.exports = {
     "productName": "BlueBubbles",
-    "appId": "com.BlueBubbles.BlueBubbles-Server",
+    "appId": "com.BlueBubbles.BlueBubbles",
     "npmRebuild": true,
     "directories": {
         "output": "releases",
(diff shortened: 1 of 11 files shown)


zlshames:
I'm a bit confused here... We don't really rely on timed polling for the iMessage DB. We leverage system file events to hook into when the DB files have been updated, and that kicks off a poll of the DB. There is a "look back" time, but it's just in case, for the first pill I think... I'll need to re look at the code to be 100% sure but that's what I remember off the top of my head

Issue #812, opened 2026-06-13
statistics/totals & /media: synchronous COUNT(*) on the main thread (better-sqlite3), no caching — blocks the event loop on large chat.db


johndkos:
## Summary

`GET /api/v1/server/statistics/totals` and `/statistics/media` compute their counts with **synchronous `better-sqlite3` queries on the main thread**, and cache nothing. On a large `chat.db`, each `COUNT(*)` takes seconds; because the driver is synchronous it **blocks the Node event loop** for that entire time, stalling every other in-flight HTTP request — including the socket health/reachability probe. Clients interpret the missed probe as the server being unreachable and tear down their socket (then reconnect, often re-triggering the same call). On a small DB it's invisible; it scales badly with message history.

## Where (refs @ `88a4921`, `development`)

The iMessage `DataSource` uses the **synchronous** `better-sqlite3` driver:
- [`databases/imessage/index.ts#L37-L42`](https://github.com/BlueBubblesApp/bluebubbles-server/blob/88a4921bbd5a8111f1e9582b83715cf877171037/packages/server/src/server/databases/imessage/index.ts#L37-L42) — `type: "better-sqlite3"`

The counts are plain `SELECT COUNT(*)` via TypeORM `.getCount()`, awaited sequentially:
- [`getAttachmentCount` / `getChatCount` / `getHandleCount` (index.ts#L774-L793)](https://github.com/BlueBubblesApp/bluebubbles-server/blob/88a4921bbd5a8111f1e9582b83715cf877171037/packages/server/src/server/databases/imessage/index.ts#L774-L793)
- [`getMessageCount` (index.ts#L471)](https://github.com/BlueBubblesApp/bluebubbles-server/blob/88a4921bbd5a8111f1e9582b83715cf877171037/packages/server/src/server/databases/imessage/index.ts#L471)
- [`getMediaCounts` (index.ts#L636)](https://github.com/BlueBubblesApp/bluebubbles-server/blob/88a4921bbd5a8111f1e9582b83715cf877171037/packages/server/src/server/databases/imessage/index.ts#L636)

`ServerInterface` awaits these directly and caches nothing:
- [`getDatabaseTotals` (serverInterface.ts#L6-L30)](https://github.com/BlueBubblesApp/bluebubbles-server/blob/88a4921bbd5a8111f1e9582b83715cf877171037/packages/server/src/server/api/interfaces/serverInterface.ts#L6-L30)
[...]

PR #813, opened 2026-06-13
fix: cache native contact lookups so the synchronous getAllContacts() can't block the event loop (#811)

### LINKED ISSUE #811: Contact fetch with extraProperties=avatar blocks the Node event loop ~5s, causing client disconnect/reconnect loops ("connecting")
**Environment**
- Server: 1.9.9 (latest stable)
- macOS: 15.7.7 (build 24G720)
- Messages: 14.0
- Private API: enabled, mode `process-dylib`, helper connected (`helper_connected: true`)
- ~2,500 contacts, served live from the macOS Contacts framework (the server's own `contact` DB table is empty)

**Summary**

`GET /api/v1/contact?extraProperties=avatar` takes ~4.3–5.7s and **blocks the Node event loop** for its duration. Clients re-request contacts-with-avatars on every (re)connect; because the block exceeds the client's ~5s reachability-probe window, the client tears down the socket (`client namespace disconnect`) and reconnects, which re-triggers the fetch. The result is a repeating loop that surfaces on clients as being stuck on "connecting."

**Reproduction** (measured on the server itself, bypassing any tunnel)

Idle `GET /api/v1/ping` ≈ 0.002s. Fire `GET /api/v1/contact?extraProperties=avatar` in the background and ping repeatedly during it:

```
ping #1 (during avatar fetch): 4.83s
ping #2–9 (after the synchronous phase): ~0.002s each
```

A single ~5s synchronous block, not a gradual slowdown. For comparison, on the same server:
- `GET /api/v1/contact` (no avatars): ~6–25ms
- `GET /api/v1/server/statistics/totals`: ~3ms

Only the avatar variant blocks.

**Server log pattern** (redacted)

```
Request to /api/v1/contact (URL Params: {"extraProperties":"avatar"})
Request to /api/v1/contact took 5313 ms
[SocketRoutes] Client <id> disconnected! Reason: client namespace disconnect
Request to /api/v1/contact (URL Params: {"extraProperties":"avatar"})
Request to /api/v1/contact took 5455 ms
[SocketRoutes] Client <id> disconnected! Reason: client namespace disconnect
```
(repeating every ~15–30s)

**Suspected cause**

Synchronous avatar loading from the macOS Contacts framework (`node-mac-contacts`) on the main thread; cost scales with contact count. At ~2,500 contacts the block sits right around the client's ~5s probe timeout, which makes the disconnects intermittent.

**Impact**

Clients (Windows desktop observed) repeatedly drop to "connecting"; each reconnect re-triggers the blocking fetch.

**Possible fixes**

- Load avatars off the main thread (worker), or stream/paginate, or cache avatars server-side so the request is a fast read rather than a live Contacts-framework call.
- Is this already addressed by the contacts API work in #701 / the `development` branch? If so, would a stable release resolve it (cf. #700 re: a macOS 15 / 2.0.0 release)?

A fuller redacted server log (last ~3h) is available if helpful.

johndkos:
## What

Cache `ContactsLib.getAllContacts()` results so repeat calls don't re-run the synchronous `node-mac-contacts` native lookup. Fixes #811.

## Why

`ContactsLib.getAllContacts()` calls `node-mac-contacts`' **synchronous** `getAllContacts(extraProps)` on every request. When images/thumbnails are requested (`extraProperties=avatar` → `contactImage`/`contactThumbnailImage`), it marshals avatar data for every contact — multi-second on large address books. Because the call is synchronous, it **blocks the Node event loop** for its full duration, stalling every other in-flight HTTP request — including the socket health/reachability probe. Clients (observed: Windows desktop) treat the missed probe as the server being unreachable, drop the socket, reconnect, and re-request contacts-with-avatars — re-triggering the block. The result is a reconnect loop that surfaces as the client stuck on "connecting".

The file already had the *intent* to cache (an `isFirstLoad` flag + a `contact-changed` listener) but never stored the loaded result, so every call re-paid the full native cost.

Measured on a live 1.9.9 server (~2,500 contacts, macOS 15.7.7, Messages 14): `GET /api/v1/contact?extraProperties=avatar` ≈ 4.3–5.7 s, and a `GET /api/v1/ping` fired during it was delayed ~4.8 s (vs ~2 ms idle), confirming the event-loop block. Plain `/contact` (no avatars) and `/server/statistics/*` were single-digit ms.

## How

- Cache the native result in a `Map` keyed by the normalized + sorted requested extra-prop set.
- Serve repeat requests from cache, keeping the synchronous native call off the hot path.
- Invalidate on the existing `contact-changed` OS listener (immediate) plus a 5-minute TTL safety net for any missed event.

A client's reconnect re-fetch now hits the cache and returns instantly, so the probe no longer times out and the loop stops. The first (cold) load still runs once; moving it off-thread (worker) is a possible follow-up, kept out of scope to stay minimal.

## Testing

The block was measured on a live 1.9.9 server as described above. The change is localized to `ContactsLib` and reuses the existing change-listener. I did not run a full end-to-end build against a production database (to avoid standing up a second server instance alongside a live one) — happy to adjust the TTL or approach per maintainer preference.

## Related

Sibling of #812 — the statistics endpoints share the same synchronous-on-main-thread + no-cache pattern (`better-sqlite3` `COUNT(*)`).

### DIFF
--- a/packages/server/src/server/api/lib/ContactsLib.ts
+++ b/packages/server/src/server/api/lib/ContactsLib.ts
@@ -28,6 +28,16 @@ export class ContactsLib {
         'socialProfiles'
     ];
 
+    // In-memory cache of native contact results, keyed by the requested extra-prop set.
+    // node-mac-contacts' getAllContacts() is SYNCHRONOUS and, when images/thumbnails are
+    // requested, marshals avatar data for every contact — multi-second on large address
+    // books, which blocks the Node event loop and stalls all other HTTP requests (including
+    // the socket health probe), causing clients to drop and reconnect. Caching makes repeat
+    // calls instant so the synchronous native call stays off the hot path.
+    private static cache: Map<string, any[]> = new Map();
+    private static cacheLoadedAt = 0;
+    private static readonly cacheTtlMs = 5 * 60 * 1000; // safety net; also cleared on OS contact-change
+
     static async requestAccess() {
         if (!contacts) return "Unknown";
         return await contacts.requestAccess();
@@ -46,6 +56,18 @@ export class ContactsLib {
     static getAllContacts(extraProps: string[] = []) {
         if (!contacts) return [];
 
+        // Drop the cache after the TTL as a safety net in case an OS change event is missed.
+        if (ContactsLib.cacheLoadedAt && Date.now() - ContactsLib.cacheLoadedAt > ContactsLib.cacheTtlMs) {
+            ContactsLib.invalidateCache();
+        }
+
+        // Serve from cache when we already have a result for this exact prop set. This keeps
+        // the synchronous native call (and its event-loop block) off the hot path — repeat
+        // requests, e.g. clients re-fetching contacts-with-avatars on every reconnect, are free.
+        const key = JSON.stringify(extraProps.map(e => e.toLowerCase()).sort());
+        const cached = ContactsLib.cache.get(key);
+        if (cached) return cached;
+
         // If it's the first load, we need to load all available info.
         // And also listen for changes so we can reload all the info again.
         if (isFirstLoad) {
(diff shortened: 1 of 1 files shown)


zlshames:
The only thing I want to be conscious of here is memory usage, especially on older macOS systems. Theoretically, any time the extraProps changes, it could fetch and cache a new, separate version of the contacts list. I'll have to think about this some more. Not a huge issue, but if someone has a huge DB of contacts, then it could be

PR #820, opened 2026-07-07
Fix NotificationCenter DB path on newer macOS

franklinho:
## Summary
- Prefer the newer macOS `group.com.apple.usernoted` NotificationCenter database path when it exists.
- Fall back to the legacy `com.apple.notificationcenter` database path.
- Await the existing async DB existence check so startup reports missing notification DBs correctly.

## Testing
- Validated locally on macOS 26 with the installed BlueBubbles server ASAR using the usernoted DB path.
- Ran `git diff --check` on the patch branch.

### DIFF
--- a/packages/server/src/server/databases/notificationCenter/NotiicationCenterRepository.ts
+++ b/packages/server/src/server/databases/notificationCenter/NotiicationCenterRepository.ts
@@ -1,4 +1,6 @@
 import fs from "fs";
+import os from "os";
+import path from "path";
 import { Brackets, DataSource } from "typeorm";
 import { Record } from "./entity/Record";
 import { App } from "./entity/App";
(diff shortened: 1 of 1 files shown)


zlshames (approved):
LGTM

Edit: nvm

zlshames:
Wait actually... can you verify one thing for me. Is the schema of the new location the same as the old one? Also, why is this still a draft?

franklinho:
yes, the new path has the same schema the server expects.
New DB path checked:
~/Library/Group Containers/group.com.apple.usernoted/db2/db
It contains:
CREATE TABLE record (
  rec_id INTEGER PRIMARY KEY,
  app_id INTEGER,
  uuid BLOB,
  data BLOB,
  request_date REAL,
  request_last_date REAL,
  delivered_date REAL,
  presented Bool,
  style INTEGER,
  snooze_fire_date REAL
);

CREATE TABLE app (
  app_id INTEGER PRIMARY KEY,
  identifier VARCHAR,
  badge INTEGER NULL
);
That matches the server’s TypeORM entities:
Record.ts maps rec_id, app_id, uuid, data, request_date, request_last_date, delivered_date, presented, style, snooze_fire_date
App.ts maps app_id, identifier, badge

PR #821, opened 2026-07-09
Add polling fallback for iMessage listener

franklinho:
## Summary

Adds a periodic polling fallback to the iMessage database listener while keeping the existing `fs.watch` path for low-latency delivery.

The listener already had a `db_poll_interval` config default, but the active listener path only reacted to filesystem change events for `chat.db` / `chat.db-wal`. On macOS, those events can be missed or delayed even when Messages has committed new rows. When that happens, BlueBubbles does not emit `new-message` and no FCM notification is sent until a client later runs an incremental sync.

## What changed

- Adds `pollIntervalMs` and a timer to `IMessageListener`.
- Reuses the existing debounced `handleChangeEvent` path so watcher events and interval polls share the same lock, lookback, cache trimming, and duplicate-event protection.
- Clears the interval and stops the watcher in `stop()`.
- Wires the interval to existing `db_poll_interval` config.

## Local verification

Reproduced on macOS with an Android BlueBubbles client:

1. A self-message appeared in `~/Library/Messages/chat.db`, but the server did not log `New Message` or send FCM until the Android app later queried `/api/v1/message/query`.
2. After this patch and repacking the local app, a new self-message produced:
   - server log: `New Message from +*******9008, "Hello"`
   - server log: `[FCMService] Sending FCM notification (Priority: high) to 1 device(s)`
   - Android log: `Received event of type new-message from Socket.io`
   - Android log: `Received new message of type new-message from FCM`
   - Android active notification on `com.bluebubbles.new_messages` with tag `com.bluebubbles.messaging.NEW_MESSAGE_NOTIFICATION`.

Build verification:

```bash
cd packages/server
npm run build
```

This completed successfully.

### DIFF
--- a/packages/server/src/server/databases/imessage/listeners/IMessageListener.ts
+++ b/packages/server/src/server/databases/imessage/listeners/IMessageListener.ts
@@ -27,19 +27,39 @@ export class IMessageListener extends Loggable {
 
     lastCheck = 0;
 
-    constructor({ filePaths, repo, cache }: { filePaths: string[], repo: MessageRepository, cache: IMessageCache }) {
+    pollIntervalMs: number;
+
+    pollTimer: NodeJS.Timeout | null = null;
+
+    constructor({
+        filePaths,
+        repo,
+        cache,
+        pollIntervalMs = 1000
+    }: {
+        filePaths: string[],
+        repo: MessageRepository,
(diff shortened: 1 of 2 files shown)


zlshames:
can you make the PR for the `development` branch

PR #825, opened 2026-07-15
Add Find My Devices support for newer macOS

qayshp:
## Summary

- Add Find My Devices support on macOS 15+ through the dedicated Find My helper.
- Keep ordinary Devices reads cache-only and make refresh an explicit API action.
- Validate helper data at the Server boundary and preserve complete versus partial snapshot semantics.

## Implementation

- `GET /api/v1/icloud/findmy/devices` returns the latest in-memory snapshot without changing Find My state.
- The helper registration hook independently warms empty Friends and Devices caches on startup.
- `POST /api/v1/icloud/findmy/devices/refresh` activates `findmy://devices`, waits for the `com.apple.findmy` helper, and requests a fresh snapshot.
- Complete snapshots replace the cache; partial snapshots merge without deleting known devices.
- Missing, malformed, duplicate, or inconsistent responses preserve the previous cache and return explicit errors.
- Normalization keeps zero battery values, rejects fabricated or invalid coordinates, and keeps direct and crowd-sourced locations distinct.
- The bundled helper is universal and verified during every Server build.

## Dependency

This draft is stacked on Friends Server PR #824. The branch is rebased directly on its current head (`e0759ee7`). The PR targets the repository's required `development` branch; once #824 merges, this PR reduces to the single Devices commit `38537e6d`.

## Validation

- `npm run test:findmy`, including cold-cache startup and atomic malformed/duplicate snapshot cases
- TypeScript `--noEmit`
- ESLint and Prettier on the review changes
- `npm run verify:findmy-helper`
- Standard minified production webpack build
- Packaged Intel app using the exact bundled helper SHA-256 `b1ee5bd7bebaf0fc091ec2c470467a9f8081ff2c33be4167ff40f6d746776ed9`
- Cold-start Devices GET before any manual refresh: HTTP 200, 37 records, 13 with direct locations, 37 unique identifiers
- Explicit Devices refresh/read: HTTP 200, 37 records, 13 with direct locations
- Explicit Friends refresh/read: HTTP 200, 8 records, 5 with coordinates
- `vmmap` confirmed the packaged helper was loaded in the live Find My process

The local smoke package used an unminified main bundle to avoid the existing development-branch TypeORM migration-name issue; the normal minified production bundle also compiled successfully.


packages/server/src/server/api/privateApi/modes/dylibPlugins/FindMyDylibPlugin.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class FindMyDylibPlugin extends DylibPlugin {
+    tag = "FindMyDylibPlugin";
+
+    parentApp = "FindMy";
+
+    bundleIdentifier = "com.apple.findmy";
+
+    get dylibPath() {
+        return path.join(FileSystem.resources, "private-api", macVer, "BlueBubblesFindMyHelper.dylib");
```
qayshp: **[blocker]** The packaged Find My helper in this PR is also x86_64-only. It cannot be injected into native arm64 Find My on Apple Silicon. Replace it with a reproducible universal artifact, update the checksum, and add an architecture check to the packaging/build workflow.

packages/server/src/server/api/interfaces/findMyInterface.ts
```diff
@@ -2,19 +2,109 @@ import { Server } from "@server";
...
+        return normalizeFindMyLocationItems(Server().findMyCache.getAll());
+    }
+
+    private static compactValue(value: any): any {
+        if (value && typeof value === "object" && "value" in value) return value.value;
+        return value;
+    }
+
+    private static finiteNumber(value: any): number | undefined {
+        const numberValue = Number(value);
```
qayshp: **[data-corruption bug]** `Number(null)` and `Number("")` both produce `0`, so an absent latitude/longitude becomes a valid-looking `(0, 0)` location; the same applies to missing numeric status fields. Reject `null`, `undefined`, and blank strings before coercion, then test those cases explicitly.

packages/server/src/server/api/interfaces/findMyInterface.ts
```diff
@@ -2,19 +2,109 @@ import { Server } from "@server";
...
+    private static transformFMIPDataManagerDevices(data: any): FindMyDevice[] {
+        const records = data?.diagnostics?.devices_data_manager?.devices_data_manager_devices?.devices_compact?.devices;
+        if (!Array.isArray(records)) return [];
+        return records.map(record => FindMyInterface.transformFMIPDeviceRecord(record));
     }
 
     static async getDevices(): Promise<Array<FindMyDevice> | null> {
         if (isMinSequoia) {
+            const papiEnabled = Server().repo.getConfig("enable_private_api") as boolean;
+            if (papiEnabled) return await FindMyInterface.refreshDevices();
```
qayshp: **[API behavior]** This turns an existing GET into a mutating, multi-second operation that launches/switches Find My and refreshes private state on every poll. Keep GET cache/snapshot-only and reserve process activation and refresh for the explicit POST endpoint; update the cache after that refresh completes.

packages/server/src/server/api/interfaces/findMyInterface.ts
```diff
@@ -65,31 +155,44 @@ export class FindMyInterface {
     }
 
     static async refreshDevices(): Promise<Array<FindMyDevice> | null> {
-        // Can't use the Private API to refresh devices yet
+        const papiEnabled = Server().repo.getConfig("enable_private_api") as boolean;
+        if (papiEnabled && isMinSequoia) {
+            checkPrivateApiStatus();
+            await FileSystem.execShellCommand("/usr/bin/open 'findmy://devices'");
+            await waitMs(3000);
```
qayshp: **[reliability blocker]** A fixed three-second sleep is not a readiness guarantee. On cold or slow startup, empty diagnostics are transformed into a successful empty list and can erase the apparent device set. Poll the helper's initialization state with a bounded timeout, distinguish unavailable/partial from a valid empty result, and preserve the last good snapshot on transient failure.

packages/server/src/server/api/interfaces/findMyInterface.ts
```diff
@@ -2,19 +2,109 @@ import { Server } from "@server";
...
+            deviceModel: rawDeviceModel,
+            rawDeviceModel,
+            deviceClass: FindMyInterface.stringValue(record?.category),
+            batteryLevel: FindMyInterface.finiteNumber(record?.batteryLevel),
+            batteryStatus: typeof record?.batteryStatus === "string" ? record.batteryStatus : "Unknown",
+            address: FindMyInterface.transformFMIPAddress(record?.address),
+            location,
+            crowdSourcedLocation: crowdSourcedLocation ?? location,
+            safeLocations: [],
+            locationEnabled: location != null,
```
qayshp: **[semantic bug]** `locationEnabled` is a device capability/setting, not an alias for `location != null`. An offline device with no current fix becomes falsely reported as disabled. Map the actual FMIP capability when available or leave this unknown; use a separate `hasLocation` concept if the API needs one. The adjacent `crowdSourcedLocation ?? location` fallback similarly relabels a direct location as crowd-sourced and should be removed.

qayshp (commented):
Maintainer review: the device pipeline proves that the data is reachable, but this is not production-ready yet. The mapper can manufacture valid-looking zero coordinates, reads have UI/process side effects, readiness is based on a fixed sleep, and several output fields change FMIP semantics. Add mapper fixtures for null/blank/numeric values, enum/status fields, timestamps, addresses, and unavailable/empty/partial responses. This PR is stacked on #824; after the Friends PR lands, rebase it so the final diff is Devices-only.

qayshp (commented):
Second maintainer pass found and fixed two merge-blocking Server issues:

1. The helper registration hook only warmed Friends, so a cold Server launch could leave the cache-only Devices GET empty until a client discovered and called the explicit refresh route. Friends and Devices caches now warm independently.
2. A complete helper response containing a mix of valid and malformed records could be normalized and then replace the cache, silently dropping the bad records. Complete and partial responses now reject any malformed returned record or duplicate identifier and preserve the last good cache.

Regression tests cover all cold-cache combinations plus malformed, mixed, and duplicate snapshots. The final package passed TypeScript, lint, formatting, helper verification, and production builds. Before any manual refresh, cold GET returned 37 Devices with 13 direct locations and 37 unique identifiers; explicit refresh returned the same result. Friends remained functional at 8 records with 5 coordinates, and the packaged helper was confirmed loaded in Find My.

I do not see another merge-blocking Server issue. The local unsigned smoke package still lacks its installed app's System Events grant, so it logs a denial when hiding windows; that does not affect helper registration or either API result. This PR remains dependent on #824.

Issue #733, opened 2025-04-19
Headless mode + Always Start via Terminal = no launch


srishChicago:
Running Bluebubbles v1.9.8 on Mac OS 13.7.5

When setting the app to "Headless mode" and in the Advanced Features Settings: "Always Start via Terminal" the app, the Bluebubbles app seems to launch, but the terminal to actually load the listening services doesn't start, which renders the app non-functional.

Logging in case other have this issue. Workaround for me is to turn off "Headless mode"

zlshames:
i think this is a known issue. I believe I've tried to resolve it in the past but had issues. I'll have to look into it again...

TurboTheTurtle:
Can confirm this also affects **v1.9.9** on macOS 26.4.0 (Apple M4).

The root cause is in `preChecks()` at `main.js:1:362243` — it calls `window.minimize()` but the window object is null in headless mode:

```
TypeError: Cannot read properties of null (reading 'minimize')
    at ap.preChecks (dist/main.js:1:362243)
    at ap.initServer (dist/main.js:1:354379)
    at async ap.start (dist/main.js:1:359795)
```

This is an unhandled promise rejection that prevents `initServer` from completing — the HTTP server never starts, so the WebUI and API are unreachable. The process stays alive but non-functional.

**Workaround:** Set `headless=0`, `start_minimized=1`, `hide_dock_icon=1` in the config DB. This creates the window (so `minimize()` does not crash) but immediately hides it. Functionally equivalent to headless mode.

If running via LaunchAgent or nohup, also redirect stdin from `/dev/null` to prevent Electron from suspending on TTY read.

PR #831, opened 2026-07-23
Fix headless startup window minimization

### LINKED ISSUE #789: Headless mode crashes with null window reference in preChecks when launched without WindowServer
## Description

When BlueBubbles is launched in headless mode (`headless: 1` in config.db) from a session without a WindowServer connection (e.g., SSH), the server crashes during initialization with an unhandled promise rejection.

## Steps to Reproduce

1. Enable headless mode in BlueBubbles settings (or set `headless|1` in config.db)
2. Close BlueBubbles
3. SSH into the Mac as the user that runs BlueBubbles
4. Launch BlueBubbles from the SSH session:
   ```bash
   nohup /Applications/BlueBubbles.app/Contents/MacOS/BlueBubbles > /tmp/bb.log 2>&1 &
   ```
5. Server starts initialization but crashes during `preChecks`

## Error Output

```
[info] Headless mode enabled, skipping window creation...
[info] [BlueBubblesServer] Running pre-start checks...
(node:4507) UnhandledPromiseRejectionWarning: TypeError: Cannot read properties of null (reading 'minimize')
    at ap.preChecks (/Applications/BlueBubbles.app/Contents/Resources/app.asar/dist/main.js:1:362243)
    at ap.initServer (/Applications/BlueBubbles.app/Contents/Resources/app.asar/dist/main.js:1:354379)
    at async ap.start (/Applications/BlueBubbles.app/Contents/Resources/app.asar/dist/main.js:1:359795)
```

The server never starts listening on the configured port (1234).

## Expected Behavior

When headless mode is enabled, `preChecks` should guard against a null window reference. The `minimize()` call should be skipped if the BrowserWindow is null, since headless mode intentionally skips window creation.

## Workaround

Launch BlueBubbles from a macOS GUI session (e.g., via Fast User Switching or a LaunchAgent that runs in the user's login session). In a GUI context, Electron creates an invisible window even in headless mode, so the `minimize()` call succeeds.

A LaunchAgent with `RunAtLoad: true` and `KeepAlive` works reliably for unattended operation.

## Environment

- BlueBubbles Server: v1.9.9
- macOS: 26.4.0 (Tahoe)
- Apple Silicon (M4)
- Headless mode: enabled
- Private API: enabled

## Related

- #733 (Headless mode + Always Start via Terminal)

qayshp:
## Summary

- Skip window minimization when headless startup intentionally has no `BrowserWindow`.
- Preserve the existing start-minimized behavior whenever a window is available.
- Add a focused regression check for the headless/null-window, disabled, and ordinary minimization paths.

## Root cause

Headless mode returns from `AppWindow.build()` without creating a `BrowserWindow`. During server initialization, `preChecks()` still reads the independent `start_minimized` setting and unconditionally calls `this.window.minimize()`. When both settings are enabled, the null window dereference rejects startup before the HTTP server begins listening.

The minimization decision now goes through a small null-safe helper. This keeps the normal GUI behavior unchanged while making the headless combination a no-op.

## User impact

BlueBubbles can start headlessly from SSH, a LaunchAgent, or another session without a WindowServer connection even when `start_minimized` remains enabled. Existing GUI launches with start-minimized enabled still minimize their window.

## Validation

Passed:

- `npm --workspace @bluebubbles/server run test:headless-window`
- Prettier check for all changed files
- Targeted ESLint for `src/server/index.ts` and `src/server/utils/WindowUtils.ts`
- Targeted TypeScript `--noEmit` check for `WindowUtils.ts`
- `git diff --check`

The full repository checks still expose unrelated errors in unchanged `origin/development` files:

- Full server lint: existing `max-len` errors in `src/server/api/privateApi/modes/dylibPlugins/index.ts` lines 143 and 158.
- Production server build: existing `NodeJS.Timer` / `clearInterval` type mismatch in `src/server/lib/ScheduledService.ts` line 39.

Fixes #789.

Related: #733.

### DIFF
--- a/packages/server/src/server/utils/WindowUtils.ts
+++ b/packages/server/src/server/utils/WindowUtils.ts
@@ -1,0 +1,7 @@
+import type { BrowserWindow } from "electron";
+
+export const minimizeWindowIfRequested = (startMinimized: boolean, window: BrowserWindow | null): void => {
+    if (startMinimized && window) {
+        window.minimize();
+    }
+};
--- a/packages/server/src/server/utils/WindowUtils.ts
+++ b/packages/server/src/server/utils/WindowUtils.ts
@@ -1,7 +1,8 @@
 import type { BrowserWindow } from "electron";
 
-export const minimizeWindowIfRequested = (startMinimized: boolean, window: BrowserWindow | null): void => {
-    if (startMinimized && window) {
-        window.minimize();
-    }
+export const minimizeWindowIfRequested = (startMinimized: boolean, window: BrowserWindow | null): boolean => {
+    if (!startMinimized || !window) return false;
+
+    window.minimize();
+    return true;
 };
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -12,6 +12,7 @@
         "build": "export NODE_ENV=production && webpack --config ./scripts/webpack.main.prod.config.js",
         "start": "export NODE_ENV=development && webpack --config ./scripts/webpack.main.config.js && electron ./dist/main.js",
         "lint": "eslint --ext=jsx,js,tsx,ts src",
+        "test:headless-window": "node ./scripts/test-headless-window.js",
         "dist": "npm run build && electron-builder build --mac --publish never --config ./scripts/electron-builder-config.js",
         "release": "npm run build && electron-builder build --mac --publish always --config ./scripts/electron-builder-config.js",
         "rebuild": "electron-rebuild -f better-sqlite3 node-mac-contacts node-mac-permissions",
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -9,10 +9,10 @@
     },
     "description": "BlueBubbles Server is the app that powers the BlueBubbles app ecosystem",
     "scripts": {
+        "test:headless-window": "node ./scripts/test-headless-window.js",
         "build": "export NODE_ENV=production && webpack --config ./scripts/webpack.main.prod.config.js",
         "start": "export NODE_ENV=development && webpack --config ./scripts/webpack.main.config.js && electron ./dist/main.js",
         "lint": "eslint --ext=jsx,js,tsx,ts src",
-        "test:headless-window": "node ./scripts/test-headless-window.js",
         "dist": "npm run build && electron-builder build --mac --publish never --config ./scripts/electron-builder-config.js",
         "release": "npm run build && electron-builder build --mac --publish always --config ./scripts/electron-builder-config.js",
         "rebuild": "electron-rebuild -f better-sqlite3 node-mac-contacts node-mac-permissions",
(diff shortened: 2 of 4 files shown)


packages/server/src/server/utils/WindowUtils.ts
```diff
@@ -0,0 +1,7 @@
+import type { BrowserWindow } from "electron";
+
+export const minimizeWindowIfRequested = (startMinimized: boolean, window: BrowserWindow | null): void => {
+    if (startMinimized && window) {
```
qayshp: [minor] When `startMinimized` is true but `window` is null, this correctly becomes a no-op, but the decision is silent. Could the helper report whether minimization occurred so `preChecks()` can debug-log the intentional headless skip? That would make future startup reports distinguish expected headless behavior from a lost setting.

qayshp (commented):
Self-review: the null guard addresses the reported crash and the focused regression check covers both headless and ordinary window paths. I found one minor diagnostics follow-up to address before this draft is marked ready.

packages/server/src/server/utils/WindowUtils.ts
qayshp: Plan: before marking this draft ready, I will have `minimizeWindowIfRequested` return whether it actually minimized a window, add a debug log in `preChecks()` when `start_minimized` is enabled but no window exists, and extend the focused test to assert both outcomes. I will then rerun the regression script, Prettier, targeted ESLint and TypeScript checks, and the production build attempt before pushing a follow-up commit and resolving this thread.

qayshp:
## Validation update

- **Passed:** `npm --workspace @bluebubbles/server run test:headless-window`
- The focused check covers a null headless window, disabled start-minimized behavior, and ordinary window minimization.

A fresh live application startup was not completed. The normal production build remains blocked by the unrelated pre-existing `ScheduledService.ts:39` `NodeJS.Timer` / `clearInterval` type error, so this result does not claim an end-to-end packaged-app pass.

packages/server/src/server/utils/WindowUtils.ts
qayshp: Implemented the review plan in ba9c7b4f: the helper now reports whether minimization occurred, the headless startup path emits a debug diagnostic when no BrowserWindow exists, and the focused regression covers missing, disabled, and real-window cases. `node packages/server/scripts/test-headless-window.js`, targeted ESLint/Prettier, and `git diff --check` pass. A full type-check still reaches the unrelated existing `ScheduledService.ts:39` Timer/Timeout error. This feedback is fully addressed.

qayshp:
Latest-head validation (`235ae213`):

- Moved only the unchanged focused-test script entry to a distinct `package.json` anchor so this PR merges cleanly with the sibling Server PRs.
- Re-ran the focused check under Node 20: passed.
- Changed-file Prettier, targeted ESLint, full-server lint comparison, and `git diff --check` behaved as documented.
- The production build reaches only the unchanged `ScheduledService.ts:39` type error already present on the base branch.
- A sequential merge simulation against current `origin/development` with Server #832–#834 completed without conflicts.

PR #832, opened 2026-07-23
Resolve Tahoe fallback service from chat metadata

### LINKED ISSUE #777: macOS 26 (Tahoe): AppleScript sendMessage fails with error -1700 due to 'any' service type
## Description

On macOS 26 (Tahoe), sending messages via the AppleScript fallback fails with:

```
error "不能将\"any\"转换为\"constant\"类型。" number -1700
```

## Root Cause

macOS 26 changed the chat GUID prefix in the Messages database from `iMessage;-;` to `any;-;`. When BlueBubbles extracts the service type from the chat GUID (e.g., `any;-;+64210766404`), it gets `"any"` instead of `"iMessage"`.

The `buildServiceScript` function in `src/server/api/apple/actions.ts` (via `scripts.ts`) generates:

```applescript
set targetService to 1st account whose service type = any
```

The Messages AppleScript dictionary does not recognize `any` as a valid service type constant, so it throws error -1700. The valid constants are `iMessage`, `SMS`, and `RCS`.

## Affected Code

`src/server/api/apple/scripts.ts`, `buildServiceScript` function:

```typescript
const buildServiceScript = (inputService: string) => {
    let theService = inputService;
    if (!isMinBigSur && theService !== "iMessage") {
        theService = `"${theService}"`;
    }
    const svcClass = isMinVentura ? 'account' : 'service';
    let serviceScript = `set targetService to 1st ${svcClass} whose service type = ${theService}`;
    // ...
};
```

When `inputService` is `"any"` (extracted from macOS 26 chat GUIDs), the generated AppleScript is invalid.

## Proposed Fix

Map `"any"` to `"iMessage"` before building the AppleScript:

```typescript
const buildServiceScript = (inputService: string) => {
    let theService = inputService;
    // macOS 26 uses "any" as chat GUID prefix instead of "iMessage"
    if (theService === "any") {
        theService = "iMessage";
    }
    // ... rest of function
};
```

This is safe because:
- `any` is not a valid AppleScript service type, so it would always fail
- Messages addressed to `any;-;` recipients are iMessage conversations
- The primary send path (using `chat id`) already works fine with `any;-;` GUIDs — only the fallback path is affected

## Environment

- macOS 26.3 (Tahoe), Apple Silicon
- BlueBubbles Server (latest from Homebrew)
- Private API: disabled (SIP enabled)

## Workaround

Patching `dist/main.js` in the ASAR to add `"any"===t&&(t="iMessage")` in the compiled `buildServiceScript` function.

qayshp:
## Summary

- resolve Tahoe `any` chat GUIDs to a concrete Messages service from the matching chat's metadata before generating the address-based AppleScript fallback
- accept only `iMessage`, `SMS`, or `RCS` as resolved services and fail closed when the lookup fails, returns no matching service, or returns an unsupported value
- keep service-lookup failure diagnostics generic instead of logging arbitrary database exception text
- preserve explicit service GUID prefixes, the unqualified-address iMessage default, and the primary chat-ID send path
- add focused regression coverage for service resolution, failure handling, and the `ActionHandler` fallback wiring

## Root cause

On macOS 26 Tahoe, `any` is a transport-agnostic chat GUID prefix observed across iMessage, SMS, and RCS chats. Messages AppleScript still requires a concrete service type, so interpolating `any` into the fallback script raises AppleScript error `-1700`. A hardcoded or unresolved `any` → `iMessage` mapping could misroute a non-iMessage conversation.

## Implementation

When the primary chat-ID send fails for an `any;...` GUID, `ActionHandler` queries the exact matching chat without loading participants and passes its `serviceName` to `sendMessageFallback()`. The script generator:

- honors explicit `iMessage`, `SMS`, and `RCS` GUID prefixes
- uses a resolved `iMessage`, `SMS`, or `RCS` value for `any` GUIDs
- throws a clear service-resolution error for an unresolved or unsupported `any` service before any fallback AppleScript is executed
- retains the existing iMessage default for unqualified addresses

If the metadata lookup itself fails, the debug log records only the generic failure decision, not the underlying exception text.

## Validation

Passed on the current head:

- `node --test test/tahoeFallbackService.test.cjs` — 6/6 tests
- targeted ESLint
- targeted Prettier for the newly changed code; the unchanged quote style in `scripts.ts` still matches the base branch
- `git diff --check`

The focused suite covers:

- resolved `any` iMessage, SMS, and RCS selection through both the script generator and `ActionHandler`
- authoritative explicit service GUIDs and proof that they bypass chat-metadata lookup
- the unqualified-address iMessage default
- lookup exceptions, empty results, null services, and unsupported services
- exact public errors, privacy-safe debug and warning logs, and proof that failure cases do not execute fallback AppleScript

Live validation on a SIP-enabled Tahoe host used the current production implementation without publishing identifiers or message content:

- a resolved `any` iMessage route recorded outgoing, sent, delivered, and no error
- a resolved `any` SMS route recorded outgoing, sent, and no error; no delivery receipt was available, so this is a dispatch pass
- a resolved `any` RCS route recorded outgoing, sent, delivered, and no error

Each live run forced the primary chat-ID path to fail, performed one exact-chat metadata lookup, executed one correctly resolved address-based fallback, and created one matching outgoing record. The failure-only cases are covered deterministically by the focused suite; no Messages database fault injection was needed. No recipient-confirmed delivery is claimed.

The production build reaches only the existing unrelated `ScheduledService.ts:39` `Timer`/`Timeout` type error.

Fixes #777

### DIFF
--- a/packages/server/src/server/api/apple/actions.ts
+++ b/packages/server/src/server/api/apple/actions.ts
@@ -91,7 +91,24 @@ export class ActionHandler {
             try {
                 // Generate the new send script
                 log.debug(`Sending AppleScript text using fallback script...`);
-                messageScript = sendMessageFallback(chatGuid, message ?? "", theAttachment);
+                let resolvedService: string = null;
+                if (chatGuid.startsWith("any;")) {
+                    try {
+                        const [chats] = await Server().iMessageRepo.getChats({
+                            chatGuid,
+                            withParticipants: false,
+                            limit: 1
+                        });
+                        resolvedService = chats[0]?.serviceName ?? null;
+                    } catch (serviceError: any) {
+                        log.debug(
+                            `Failed to resolve the fallback chat service: ` +
+                                `${serviceError?.message ?? String(serviceError)}`
+                        );
+                    }
+                }
+
+                messageScript = sendMessageFallback(chatGuid, message ?? "", theAttachment, resolvedService);
                 await FileSystem.executeAppleScript(messageScript);
             } catch (ex: any) {
                 error = ex;
--- a/packages/server/src/server/api/apple/actions.ts
+++ b/packages/server/src/server/api/apple/actions.ts
@@ -100,11 +100,8 @@ export class ActionHandler {
                             limit: 1
                         });
                         resolvedService = chats[0]?.serviceName ?? null;
-                    } catch (serviceError: any) {
-                        log.debug(
-                            `Failed to resolve the fallback chat service: ` +
-                                `${serviceError?.message ?? String(serviceError)}`
-                        );
+                    } catch {
+                        log.debug("Failed to resolve the fallback chat service; using the legacy iMessage default.");
                     }
                 }
 
--- a/packages/server/src/server/api/apple/actions.ts
+++ b/packages/server/src/server/api/apple/actions.ts
@@ -91,7 +91,7 @@ export class ActionHandler {
             try {
                 // Generate the new send script
                 log.debug(`Sending AppleScript text using fallback script...`);
-                let resolvedService: string = null;
+                let resolvedService: string | null = null;
                 if (chatGuid.startsWith("any;")) {
                     try {
                         const [chats] = await Server().iMessageRepo.getChats({
@@ -101,7 +101,9 @@ export class ActionHandler {
                         });
                         resolvedService = chats[0]?.serviceName ?? null;
                     } catch {
-                        log.debug("Failed to resolve the fallback chat service; using the legacy iMessage default.");
+                        log.debug(
+                            "Failed to resolve the fallback chat service; the fallback message will not be sent."
+                        );
                     }
                 }
 
--- a/packages/server/src/server/api/apple/scripts.ts
+++ b/packages/server/src/server/api/apple/scripts.ts
@@ -64,8 +64,9 @@ const getServiceFromInput = (value: string) => {
     // so we should default to iMessage
     if (valSplit.length <= 1) return "iMessage";
 
-    // Otherwise, return the "first" index in the array,
-    return valSplit[0];
+    // Tahoe uses "any" for iMessage chat GUIDs, but Messages AppleScript requires a concrete service type.
+    const service = valSplit[0];
+    return service === "any" ? "iMessage" : service;
 };
 
 /**
--- a/packages/server/src/server/api/apple/scripts.ts
+++ b/packages/server/src/server/api/apple/scripts.ts
@@ -15,7 +15,7 @@ const buildServiceScript = (inputService: string) => {
         theService = `"${theService}"`;
     }
 
-    const svcClass = isMinVentura ? "account" : "service";
+    const svcClass = isMinVentura ? 'account' : 'service';
 
     let serviceScript = `set targetService to 1st ${svcClass} whose service type = ${theService}`;
     if (!isMinBigSur && theService !== "iMessage") {
--- a/packages/server/src/server/api/apple/scripts.ts
+++ b/packages/server/src/server/api/apple/scripts.ts
@@ -56,7 +56,7 @@ const getAddressFromInput = (value: string) => {
     return valSplit[valSplit.length - 1];
 };
 
-const getServiceFromInput = (value: string, resolvedService: string = null) => {
+const getServiceFromInput = (value: string, resolvedService: string | null = null) => {
     // This should always produce an array of minimum length, 1
     const valSplit = value.split(";");
 
@@ -69,7 +69,9 @@ const getServiceFromInput = (value: string, resolvedService: string = null) => {
     if (service !== "any") return service;
 
     const supportedServices = ["iMessage", "SMS", "RCS"];
-    return supportedServices.includes(resolvedService) ? resolvedService : "iMessage";
+    if (resolvedService && supportedServices.includes(resolvedService)) return resolvedService;
+
+    throw new Error("Unable to resolve a supported Messages service for this chat; the fallback message was not sent.");
 };
 
 /**
@@ -205,7 +207,7 @@ export const sendMessageFallback = (
     chatGuid: string,
     message: string,
     attachment: string,
-    resolvedService: string = null
+    resolvedService: string | null = null
 ) => {
     if (!chatGuid || (!message && !attachment)) return null;
 
--- a/packages/server/src/server/api/apple/scripts.ts
+++ b/packages/server/src/server/api/apple/scripts.ts
@@ -15,7 +15,7 @@ const buildServiceScript = (inputService: string) => {
         theService = `"${theService}"`;
     }
 
-    const svcClass = isMinVentura ? 'account' : 'service';
+    const svcClass = isMinVentura ? "account" : "service";
 
     let serviceScript = `set targetService to 1st ${svcClass} whose service type = ${theService}`;
     if (!isMinBigSur && theService !== "iMessage") {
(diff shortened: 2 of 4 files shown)


packages/server/src/server/api/apple/scripts.ts
```diff
@@ -64,8 +64,9 @@ const getServiceFromInput = (value: string) => {
     // so we should default to iMessage
     if (valSplit.length <= 1) return "iMessage";
 
-    // Otherwise, return the "first" index in the array,
-    return valSplit[0];
+    // Tahoe uses "any" for iMessage chat GUIDs, but Messages AppleScript requires a concrete service type.
+    const service = valSplit[0];
+    return service === "any" ? "iMessage" : service;
```
qayshp: Self-review: this exact mapping preserves explicit `SMS` and `RCS` prefixes, but it still assumes every Tahoe `any;-;` fallback target is an iMessage chat. The generated-script assertions prove the normalization is narrow; they do not prove that Tahoe never uses `any` for a conversation that should fall back through SMS or RCS. Please keep this draft until that assumption is confirmed with real fallback sends, or replace the mapping with a service-resolution strategy if the assumption is false.

qayshp (commented):
Self-review focused on the remaining service-selection assumption and its coverage.

packages/server/src/server/api/apple/scripts.ts
qayshp: Plan before marking ready: (1) force the AppleScript address fallback on a Tahoe host for a known `any;-;` iMessage chat and record successful delivery; (2) exercise explicit `SMS;-;` and `RCS;-;` fallback GUIDs to verify account selection and delivery remain unchanged; and (3) inspect the affected Tahoe chat service fields. If `any` can represent a non-iMessage transport, I will replace the hardcoded mapping with service resolution from chat/account context and add that case to the regression assertions. The PR will remain draft until this evidence is attached.

qayshp:
## Validation update

- **Passed:** generated-AppleScript assertions for normalizing the Tahoe `any` service prefix to `iMessage`.
- Explicit `iMessage`, `SMS`, and `RCS` prefixes remained unchanged, an unqualified address retained the existing `iMessage` default, and the primary chat-ID script retained its original GUID.
- **Passed read-only:** Messages exposed an iMessage account to AppleScript; the account identifier was discarded and no send command was issued.

No message was sent. The exact address-fallback path supports only one-to-one chats and rejects group chats, so it cannot be exercised in the user-designated two-recipient conversation without testing a different code path. This remains an end-to-end validation gap.

packages/server/src/server/api/apple/scripts.ts
qayshp: The Tahoe service-field check disproved the original assumption: privacy-minimized metadata showed that `any` identifies both iMessage and RCS one-to-one chats. I replaced the hardcoded mapping in 4fc993f1 with lookup of the matching chat service, restricted the resolved value to iMessage/SMS/RCS, and added 4 regression cases including ActionHandler wiring. Focused tests, targeted ESLint/Prettier, and `git diff --check` pass. Live Tahoe sends then recorded successful delivery for resolved `any` iMessage and resolved `any` RCS routes. The explicit SMS attempt succeeded but Messages recorded the transport as RCS on that dual-capable target, so I am leaving this thread open until a true SMS-only smoke test is completed.

qayshp: Completed the remaining explicit SMS smoke test on Tahoe using an authorized SMS-only target. AppleScript returned successfully, and the privacy-minimized database check recorded service=SMS, is_from_me=1, is_sent=1, and error=0. The delivery-receipt flag remained 0, so this is claimed as successful SMS dispatch rather than recipient-confirmed delivery. Together with the previously recorded resolved `any` iMessage and resolved `any` RCS sends, all three fallback service paths now have live transport evidence. This feedback is fully addressed.

qayshp:
Superseding the earlier validation note: the final implementation no longer hardcodes Tahoe `any` GUIDs to iMessage. It resolves the matching chat's service, allow-lists `iMessage`/`SMS`/`RCS`, and retains the legacy iMessage default only when resolution is unavailable. The focused suite passes 4/4 on the current head. Live Tahoe fallback sends recorded correct iMessage and RCS delivery; the SMS-only route recorded service=SMS, outgoing/sent=true, and error=0. No SMS delivery receipt was available, so recipient-confirmed SMS delivery is not claimed.

qayshp:
Latest-head validation (`bf9e8d20`):

- Removed unrelated quote-style churn from `scripts.ts`.
- Service-lookup failures now log only the generic fallback decision; the regression check proves arbitrary exception text is not emitted and the legacy iMessage default is still used.
- The focused suite passes 4/4 under Node 20; targeted ESLint, changed-file Prettier, and `git diff --check` pass. The unchanged `scripts.ts` quote style matches the base branch.
- The production build reaches only the unchanged `ScheduledService.ts:39` type error already present on the base branch.
- A sequential merge simulation against current `origin/development` with Server #831, #833, and #834 completed without conflicts.

The previously recorded privacy-minimized live iMessage, RCS, and SMS dispatch evidence remains applicable because these follow-ups do not change service selection.

ruinthemagic (commented):
Independent validation on a SIP-enabled macOS 26.5.2 host confirms that the Tahoe any GUID prefix is transport-agnostic: it occurs on both SMS and iMessage chats. No identifiers or message content were inspected. Exact chat GUIDs are unique here, and every any-prefixed row inspected has a supported, non-null service_name, so resolving the transport from the exact chat metadata is the correct design. The focused 4-case suite and targeted ESLint pass locally; the production build reaches only the unrelated pre-existing ScheduledService Timer/Timeout error. One remaining concern: when resolution fails or returns an unsupported value, this PR defaults to iMessage. That is not truly legacy behavior for any GUIDs—the current code errors—and it could silently switch an SMS or RCS conversation to iMessage. Please consider failing closed with a clear service-resolution error instead. That should preserve transport correctness without affecting normal Tahoe chats, whose metadata resolves successfully.

packages/server/src/server/api/apple/actions.ts
```diff
@@ -91,7 +91,23 @@ export class ActionHandler {
             try {
                 // Generate the new send script
                 log.debug(`Sending AppleScript text using fallback script...`);
```
qayshp, line 93: Non-blocking: service resolution can now fail before a fallback script is generated or executed, so this log may say the fallback is being sent when none will be. Consider wording it as 'Preparing AppleScript text fallback...' so the logs distinguish preparation from execution.

qayshp (commented):
Full-diff self-review found no blocking correctness changes. One non-blocking coverage note: the focused 6-case suite is exposed as a package script but is not invoked by a root test command or GitHub Actions, so regression coverage remains manual. I would prefer wiring it into an appropriate macOS CI job when the repository has one, but I would not block this PR on introducing new CI infrastructure.

coreyhauer:
Reproduced error -1700 on BlueBubbles Server 1.9.9, macOS 26.5, Mac Studio, with Private API disconnected. Sending from Apple Messages worked. A local patch mapping `any` to `iMessage` restored sending in the affected conversation. This confirms the reported failure and workaround; I haven’t tested this PR’s implementation or SMS/RCS behavior.

jattree:
Tested this PR (head `b51c296f`) live on **macOS 26.6.2 (Tahoe, 25G83), Apple Silicon (M2), with the Private API disabled** (SIP happens to be off on this Mac, but no Private API helper was loaded during the test).

Setup: `development` @ e2a5bf6e with this PR merged, plus the two startup fixes from #840 (needed for the packaged build to start). Every chat on this Mac uses the `any;` GUID prefix (iMessage, SMS and RCS alike).

Result: `POST /api/v1/message/text` with `method: "apple-script"` to a 1:1 iMessage chat (`any;-;+1…`) returned 200 "Message sent!". Messages recorded it as `service=iMessage`, `is_sent=1`, `is_delivered=1`, `error=0`, and the server picked up the sent message. No `-1700` error.

The regression suite from this PR also passes locally (`node --test test/tahoeFallbackService.test.cjs`, 6/6, Node 20.11.1). I didn't test SMS/RCS fallback routing or group chats.

PR #833, opened 2026-07-23
Add typed, privacy-safe Google contact sync logging

qayshp:
## Summary

- report succeeded, skipped, failed, and total counts after Google contact sync
- add per-contact debug context using only the one-based batch index and field-presence booleans
- represent known contact failures with typed `ContactInterfaceError` codes
- accept only the fixed `CONTACT_ERROR_CODES` allow-list in logs; collapse all arbitrary messages and unrecognized codes to `unknown-error`
- add focused privacy and result-formatting tests

## Why

Issue #677 asks for enough detail to understand partial Google contact imports. The previous implementation tracked only whether any contact failed and logged the raw exception message. That made the final warning difficult to act on and could expose contact or provider data.

## Implementation

The final sync summary distinguishes contacts that were saved, skipped because Google supplied no name record, or failed during avatar loading or persistence. Failure debug context contains only:

- the contact's one-based position in the fetched batch
- booleans describing whether supported fields were present
- a stable failure reason from the typed contact-error allow-list, or `unknown-error`

It does not log names, phone numbers, email addresses, avatar URLs, Google resource identifiers, provider payloads, arbitrary exception messages, or merely string-shaped error codes.

## Validation

Passed on the current head:

- `node --test test/contactSyncLogging.test.cjs` — 4/4 tests
- targeted ESLint
- `git diff --check`

The tests cover aggregate formatting, present and empty field context, every allow-listed contact error, rejection of arbitrary codes, and suppression of provider/contact values and exception text.

No live Google OAuth/contact import was performed, so validation is limited to the focused automated coverage.

Fixes #677

PR #834, opened 2026-07-23
Restore legacy macOS update manifest generation

### LINKED ISSUE #817: Still include latest-mac.yml in releases at least one last time
It has been well over a year like this now, so I'm not sure how important this would be. That said, I installed Blue Bubbles on a less-frequently touched Mac that I use primarily for BlueBubbles, which is perhaps common. Since it was installed manually and is on an infrequently otherwise used machine, the only update checks were either automatic or manually either through the API from the mobile app or via the menu bar. 

I think it might be worth including a `latest-mac.yml` in at least one newer release.

Including a `latest-mac.yml` file in releases, while no longer relevant to the current update system, would allow older versions to at least detect and attempt to update. The update would fail, but at least users would know that an update was available.

References to: [#653](https://github.com/BlueBubblesApp/bluebubbles-server/issues/653#issuecomment-4826918593) & a72ef64688ec2d231faf0b4cce3523d37469bf87

qayshp:
## Summary

- re-enable electron-builder update metadata for macOS DMG releases
- restore generation and publication eligibility of `latest-mac.yml` without changing the current updater implementation

## Root cause

Commit `a72ef646` disabled `dmg.writeUpdateInfo`, so release packaging stopped producing the legacy manifest that older BlueBubbles servers still check.

## User impact

Older installations can discover that a newer release exists. Their legacy updater may still be unable to complete the upgrade, but the manifest restores the update notification path requested in the issue.

## Validation

Passed:

- loaded `electron-builder-config.js` under Node 20 and asserted:
  - `dmg.writeUpdateInfo === true`
  - the release target still includes x64 and arm64 DMGs
  - the configured publisher remains the BlueBubbles GitHub draft release
- a non-publishing electron-builder 24.13.3 fixture using this configuration produced both architecture DMGs plus `latest-mac.yml`
- the generated manifest listed both DMGs, and each recorded size and SHA-512 matched its artifact
- inspected the locked electron-builder 24.13.3 path:
  - a DMG with update info queues an update-info task
  - `writeUpdateInfoFiles()` combines both architecture entries
  - the resulting `latest-mac.yml` is dispatched as a publish artifact with the same GitHub publish configuration
- confirmed the repository release workflow calls electron-builder with `--publish always` and `GH_TOKEN`
- `git diff --check`

## Release verification

No GitHub release was created or modified for this PR. Exercising the final upload would require the upstream release credentials and would create or alter the configured BlueBubbles draft release. Because the generated manifest enters electron-builder's existing artifact publisher—the same release mechanism already responsible for the DMGs—that external upload is an operational check for the next maintainer release, not a code-level merge blocker.

After merge, the release maintainer should still confirm that `latest-mac.yml` appears beside both DMGs in the next macOS draft release.

Fixes #817

### DIFF
--- a/packages/server/scripts/electron-builder-config.js
+++ b/packages/server/scripts/electron-builder-config.js
@@ -56,7 +56,7 @@ module.exports = {
     },
     "dmg": {
         "sign": false,
-        "writeUpdateInfo": false
+        "writeUpdateInfo": true
     },
     // "afterSign": "./scripts/notarize.js"
 };


qayshp (commented):
Self-review: the configuration restores the legacy manifest switch, but the release artifact still needs end-to-end confirmation.

packages/server/scripts/electron-builder-config.js
```diff
@@ -56,7 +56,7 @@ module.exports = {
     },
     "dmg": {
         "sign": false,
-        "writeUpdateInfo": false
+        "writeUpdateInfo": true
```
qayshp, line 59: The config assertion proves this flag is enabled, but not that the release workflow uploads `latest-mac.yml` alongside both architecture artifacts. Please verify the next release dry run or workflow artifacts before marking this ready.

qayshp: Plan: before marking this ready, I will run the macOS release job in a maintainer-capable environment, inspect the packaged output and uploaded draft-release assets for `latest-mac.yml`, and verify its version, URL, and checksum fields against the generated DMGs. If it is not uploaded, I will adjust the artifact/publish configuration and repeat the check. I am leaving this thread open until that end-to-end validation is recorded.

qayshp:
## Validation update

- **Passed:** loading `electron-builder-config.js` and asserting `dmg.writeUpdateInfo === true`.
- **Passed:** a non-publishing local electron-builder 24.13.3 fixture using this PR's config produced both x64 and arm64 DMGs plus `latest-mac.yml`.
- The manifest listed both DMG files, and every recorded size and SHA-512 matched the corresponding artifact.

The fixture used `--publish never`, so nothing was uploaded and no GitHub release was created or changed. The repository's normal full build remains blocked by the unrelated pre-existing `ScheduledService.ts:39` `NodeJS.Timer` / `clearInterval` type error; release-upload validation still belongs on an actual maintainer release.

qayshp:
Merge-readiness validation:

- Node 20 config assertions confirm update metadata is enabled, both DMG architectures remain targeted, and GitHub draft publishing remains configured.
- The recorded non-publishing electron-builder 24.13.3 fixture produced both DMGs plus a two-entry `latest-mac.yml` whose sizes and SHA-512 values matched.
- Inspection of the locked builder shows that DMG update metadata is combined and dispatched with the same GitHub publish configuration; the repository release workflow invokes that publisher with `--publish always` and `GH_TOKEN`.
- `git diff --check` passes, and a sequential merge simulation against current `origin/development` with Server #831–#833 completed without conflicts.

No release was created or modified. The next maintainer release should confirm the uploaded asset, but that credentialed operational check is not a blocker for this one-line configuration change.

packages/server/scripts/electron-builder-config.js
qayshp: Re-evaluated this gate against the repository workflow and the locked electron-builder 24.13.3 implementation. The non-publishing fixture already proved that this configuration produces both architecture DMGs and a combined `latest-mac.yml`, with matching sizes and SHA-512 values. In the builder publish path, an update-enabled DMG queues update-info work, `writeUpdateInfoFiles()` combines the architecture entries, and the generated manifest is dispatched with the same GitHub publish configuration; this repository's release command uses `--publish always` with `GH_TOKEN`.

The remaining literal upload would require maintainer credentials and would create or modify the configured upstream draft release. I have therefore documented it as a post-merge operational verification for the next maintainer release, not as a code-level merge blocker. The PR description now records both the evidence and that boundary. This feedback is fully addressed.

PR #835, opened 2026-07-23
Prevent duplicate reconciled message events with a keyed cache

### LINKED ISSUE #765: [Potential Bug] Re-emitting old messages a new-message events
i think i potentially see a bug with sending new-message notifications multiple times:

- We have a cache that gets maintained that tracks "emitted" events.
- The cache gets pruned every so often, removing things that have been in the cache for > 1 hour
- When we poll the database, we look back 1 weeks time, and rely on the cache to prevent duplicates from being emitted
- Since the cache gets pruned, it allows events to be re-emitted

It's 100% weird though that people are not reporting duplicate notifications (on the Android side). I would think that they would report it if the bug was as severe as it seems to be. So maybe there is something else at play.

I know that the android app tracks GUIDs for notified messages, however, this is mainly just an in-memory cache and would not persist when the app fully closes, which probably happens daily. So this bug _should_ show itself. But it doesn't seem to.

Dupe emit bug seems to happen on macOS Sequoia. Likely happens on other versions as well.

qayshp:
## Summary

- retain message GUIDs and message state for the seven-day reconciliation window plus one reconciliation interval
- use a keyed, `Map`-backed cache for high-volume message-event deduplication
- keep chat event/state retention at one hour and preserve the existing array-backed chat cache
- centralize the message poller's reconciliation and retention constants
- add focused regression coverage for retention, pruning, and cache separation

## Root cause

The message poller periodically widens its query to seven days, but message event and state entries were pruned after one hour. A previously emitted row could therefore lose its GUID and state, then be classified as a new entry when the reconciliation lookback revisited it.

Simply extending the old array-backed event cache would also make high-volume message lookup increasingly expensive. The final implementation separates the two workloads:

- message events use `KeyedEventCache`, backed by `Map` for keyed add/find/remove operations
- chat events continue to use the existing `EventCache` array and one-hour retention
- message event and state entries remain for seven days plus five minutes, covering one final reconciliation after they age out of the query window

## Impact

Long-running servers no longer re-emit unchanged reconciled rows as `new-entry` events after the former one-hour deduplication window. Message lookup no longer performs a linear scan of the retained event set, while chat read-state deduplication keeps its existing storage and retention behavior.

## Validation

Passed on the current head:

- `node --test test/messageCacheRetention.test.cjs` — 4/4 tests
- targeted ESLint
- targeted Prettier check
- `git diff --check`

The tests cover full-lookback deduplication, expiry pruning, one-hour chat retention, `Map`-backed message storage, unchanged array-backed chat storage, and duplicate-key behavior.

`tsc -p packages/server/tsconfig.json --noEmit` still reaches the existing unrelated `ScheduledService.ts:39` `NodeJS.Timer`/`clearInterval` type error.

Fixes #765

### DIFF
--- a/packages/server/src/server/databases/imessage/pollers/constants.ts
+++ b/packages/server/src/server/databases/imessage/pollers/constants.ts
@@ -1,0 +1,10 @@
+export const MESSAGE_FAST_LOOKBACK_MS = 1000 * 60 * 30;
+
+export const MESSAGE_RECONCILE_LOOKBACK_MS = 1000 * 60 * 60 * 24 * 7;
+
+export const MESSAGE_RECONCILE_INTERVAL_MS = 1000 * 60 * 5;
+
+// Keep GUIDs through one final reconciliation after they age out of the query window.
+export const MESSAGE_DEDUP_CACHE_RETENTION_MS = MESSAGE_RECONCILE_LOOKBACK_MS + MESSAGE_RECONCILE_INTERVAL_MS;
+
+export const CHAT_DEDUP_CACHE_RETENTION_MS = 1000 * 60 * 60;
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -12,6 +12,7 @@
         "build": "export NODE_ENV=production && webpack --config ./scripts/webpack.main.prod.config.js",
         "start": "export NODE_ENV=development && webpack --config ./scripts/webpack.main.config.js && electron ./dist/main.js",
         "lint": "eslint --ext=jsx,js,tsx,ts src",
+        "test:message-cache": "node --test test/messageCacheRetention.test.cjs",
         "dist": "npm run build && electron-builder build --mac --publish never --config ./scripts/electron-builder-config.js",
         "release": "npm run build && electron-builder build --mac --publish always --config ./scripts/electron-builder-config.js",
         "rebuild": "electron-rebuild -f better-sqlite3 node-mac-contacts node-mac-permissions",
--- a/packages/server/src/server/databases/imessage/pollers/MessagePoller.ts
+++ b/packages/server/src/server/databases/imessage/pollers/MessagePoller.ts
@@ -2,7 +2,7 @@ import { IMessagePollResult, IMessagePollType, IMessagePoller } from ".";
 import { Message } from "../entity/Message";
 import { isEmpty } from "@server/helpers/utils";
 import { Server } from "@server";
-
+import { MESSAGE_FAST_LOOKBACK_MS, MESSAGE_RECONCILE_INTERVAL_MS, MESSAGE_RECONCILE_LOOKBACK_MS } from "./constants";
 
 export class MessagePoller extends IMessagePoller {
     tag = "MessagePoller";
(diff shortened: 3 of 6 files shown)


packages/server/src/server/databases/imessage/pollers/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const MESSAGE_FAST_LOOKBACK_MS = 1000 * 60 * 30;
+
+export const MESSAGE_RECONCILE_LOOKBACK_MS = 1000 * 60 * 60 * 24 * 7;
+
+export const MESSAGE_RECONCILE_INTERVAL_MS = 1000 * 60 * 5;
+
+// Keep GUIDs through one final reconciliation after they age out of the query window.
+export const MESSAGE_DEDUP_CACHE_RETENTION_MS = MESSAGE_RECONCILE_LOOKBACK_MS + MESSAGE_RECONCILE_INTERVAL_MS;
```
qayshp, line 8: Keeping message GUIDs for seven days plus one reconciliation interval is necessary for correctness, but `EventCache` currently stores a linear array. On unusually high-volume servers this increases both retained entries and lookup cost, so we should validate cache size and lookup behavior against a busy database before promoting the draft.

qayshp (commented):
Self-review complete. The cache-retention behavior is covered by the synthetic replay tests; I left one performance-oriented follow-up inline.

packages/server/src/server/databases/imessage/pollers/constants.ts
qayshp: Plan: I’ll measure `messageEvents` size and lookup time during a seven-day high-volume replay before marking the PR ready. If growth is material, I’ll replace only the message-event cache with a keyed Map/Set-backed implementation while preserving the retention window and regression coverage; the one-hour chat cache will remain unchanged.

qayshp:
## Validation update

Validated on SIP-enabled arm64 macOS 26.5.2 with Node 20.11.1 and npm 10.2.4.

- `npm run test:message-cache --workspace=@bluebubbles/server` — 3/3 tests passed.
- A synthetic benchmark using this PR's actual `EventCache` retained about 8.97 MiB at 100,000 entries; 100 worst-case lookups took 68–79 ms, and inserting those entries through `add()` took 17.18 seconds.
- At 500,000 retained entries, the cache used about 36.62 MiB and 100 worst-case lookups took about 400 ms.

The functional regression coverage passes, but the benchmark confirms that the linear cache can become material at unusually high volume. Validation against a representative busy server, or a decision to use a keyed cache, remains open before this draft is promoted.

No messages were sent or received during validation. No screenshot is needed because this change has no visual surface.

packages/server/src/server/databases/imessage/pollers/constants.ts
qayshp: Implemented the review plan in fe41ef53: message-event deduplication now uses a keyed Map-backed cache for constant-time add/find/remove while chat-event storage remains unchanged. The focused suite passes all 4 retention and duplicate-semantics cases; targeted ESLint/Prettier and `git diff --check` pass. The only full type-check failure is the unrelated existing `ScheduledService.ts:39` Timer/Timeout error. This feedback is fully addressed.

qayshp:
Superseding the earlier performance note: its linear-cache benchmark describes the prior implementation, not the current head. Message-event deduplication now uses a keyed `Map`-backed cache, while the chat cache remains the existing array-backed `EventCache`. The focused retention/cache suite passes 4/4. The remaining validation limitation is the pre-existing unrelated `ScheduledService.ts:39` `NodeJS.Timer`/`clearInterval` type-check failure.

PR #838, opened 2026-07-23
Retry zrok token setup after stale-environment cleanup

### LINKED ISSUE #678: Zrok implementation not working
I don't know if it is me. but when adding Zrok as a proxy the server address stays locked on the default cloudflare address. 
The server then reports the following error:
[ZrokService] Failed to restart Zrok! Error: [ERROR]: unable to create share (unable to create share: [POST /share][401] shareUnauthorized)
I tried the pre-release version and the latest stable, uninstall/reinstall/reboot/reconnect.
Thanks!

qayshp:
## Summary

- detect zrok's exact `you already have an enabled environment` failure when setting an auth token
- disable that stale local environment and retry `zrok enable <token>` exactly once
- stop immediately if cleanup fails and preserve the existing disable error
- keep existing invalid-token and generic-error behavior for both the initial attempt and retry
- add focused `ProcessSpawner.executeCommand` tests for the bounded recovery paths

## Root cause

`ZrokManager.setToken()` recognized the stale-environment error but returned its output as though token setup had succeeded. The stale environment remained enabled, so later share creation could continue failing. The issue's manual workaround—running `zrok disable` before enabling through BlueBubbles—showed the required recovery sequence.

## Behavior and safety

Only output containing the existing exact zrok condition triggers recovery. The sequence is bounded to:

1. `zrok enable <token>`
2. `zrok disable` if the stale-environment condition is returned
3. one final `zrok enable <token>` attempt

There is no recursion or unbounded retry. If `disable` fails, recovery stops before a second enable and preserves `Failed to disable Zrok tunnel!`. Invalid tokens remain `Invalid Zrok token!`; unrelated failures retain the existing generic error.

## Validation

Validated on current head `0cd1cb84`:

- `npm run test:zrok --workspace packages/server` — 7/7 tests passed
- the committed seventh test verifies that a failed disable stops at `enable → disable`, performs no second enable, and preserves the exact existing disable error
- targeted ESLint, Prettier, and `git diff --check` passed

## Limitation

No recovery was run against a real disposable zrok token/environment. The command sequence and errors are validated at the mocked `ProcessSpawner` boundary; this PR does not claim an end-to-end live zrok recovery pass.

The broader server type-check still reaches the pre-existing unrelated `ScheduledService.ts:39` `NodeJS.Timer`/`clearInterval` error.

Fixes #678

### DIFF
--- a/packages/server/test/zrokManager.test.cjs
+++ b/packages/server/test/zrokManager.test.cjs
@@ -1,0 +1,141 @@
+const assert = require("node:assert/strict");
+const path = require("node:path");
+const test = require("node:test");
+const babel = require("@babel/core");
+
+const loadTypeScriptModule = (modulePath, overrides = {}) => {
+    const transformed = babel.transformFileSync(modulePath, {
+        presets: [
+            [require.resolve("@babel/preset-env"), { targets: { node: "20" }, modules: "commonjs" }],
+            require.resolve("@babel/preset-typescript")
+        ]
+    });
+    const loadedModule = { exports: {} };
+    const loadModule = new Function("module", "exports", "require", transformed.code);
+    const customRequire = request =>
+        Object.prototype.hasOwnProperty.call(overrides, request) ? overrides[request] : require(request);
+    loadModule(loadedModule, loadedModule.exports, customRequire);
+    return loadedModule.exports;
+};
+
+const zrokManagerPath = path.join(__dirname, "../src/server/managers/zrokManager/index.ts");
+const token = "test-token";
+const alreadyEnabledError = {
+    output: "[ERROR]: unable to enable environment (you already have an enabled environment)"
+};
+
+const loadZrokManager = responses => {
+    const calls = [];
+    const logs = {
+        debug: [],
+        error: [],
+        info: []
+    };
+    const executeCommand = async (...args) => {
+        calls.push(args);
+        const response = responses.shift();
+        if (!response) throw new Error("Unexpected ProcessSpawner call");
+        if ("error" in response) throw response.error;
+        return response.output;
+    };
+    const logger = {
+        debug: (...args) => logs.debug.push(args),
+        error: (...args) => logs.error.push(args),
+        info: (...args) => logs.info.push(args)
+    };
+
+    const { ZrokManager } = loadTypeScriptModule(zrokManagerPath, {
+        axios: { post: async () => undefined },
+        child_process: { spawn: () => undefined },
+        electron: { app: { getVersion: () => "test" } },
+        "@server": { Server: () => undefined },
+        "@server/fileSystem": { FileSystem: { resources: "/resources" } },
+        "@server/helpers/utils": {
+            isEmpty: value => value == null || value.length === 0,
+            isNotEmpty: value => value != null && value.length > 0
+        },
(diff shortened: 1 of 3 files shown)


packages/server/test/zrokManager.test.cjs
```diff
@@ -0,0 +1,141 @@
...
+    const { calls, ZrokManager } = loadZrokManager([
+        { error: alreadyEnabledError },
+        success("environment disabled"),
+        success("environment enabled")
+    ]);
+
+    assert.equal(await ZrokManager.setToken(token), "environment enabled");
+    assert.deepEqual(calls, [
+        expectedCall(ZrokManager, ["enable", token]),
+        expectedCall(ZrokManager, ["disable"]),
```
qayshp, line 106: The command-order coverage is strong for a successful disable and a failed second enable, but the new intervening `disable` failure path is not pinned down. Please add a queued `ProcessSpawner` rejection for `disable`, assert the sequence stops at `enable -> disable` with no retry, and verify the existing `Failed to disable Zrok tunnel!` error is preserved.

qayshp (commented):
Self-review found one focused follow-up for the newly introduced recovery path.

packages/server/test/zrokManager.test.cjs
qayshp: Plan: add a seventh Node test that queues the stale-environment error followed by a failing `disable`, assert the surfaced `Failed to disable Zrok tunnel! Please check your server logs for more information.` message, and assert the command list contains only the initial `enable` and `disable`. Then rerun `test:zrok`, changed-source ESLint, Prettier, and `git diff --check`. This thread will stay open until that test is implemented and pushed.

qayshp:
## Validation update

Validated the mocked command paths on SIP-enabled arm64 macOS 26.5.2 with Node 20.11.1 and npm 10.2.4.

- `npm run test:zrok --workspace=packages/server` — 6/6 committed tests passed.
- A separate no-file-change test harness exercised a failing intervening `disable`: recovery stopped after `enable → disable`, made no second enable attempt, and preserved the existing `Failed to disable Zrok tunnel!` error.

That failure case was not added to the committed test suite, so the review follow-up remains open. Live recovery is deferred because this machine has no enabled zrok environment. Reproducing it requires a disposable zrok token/environment and will change both local and external zrok state; SIP is not the blocker.

No messages were sent or received during validation. No screenshot is needed because the command sequence and resulting state are the relevant evidence.

packages/server/test/zrokManager.test.cjs
qayshp: Implemented the requested seventh case in 7bfb0437. A disable rejection now proves recovery stops after enable -> disable, performs no second enable, and preserves the exact public disable error. All 7 focused zrok-manager cases and `git diff --check` pass. This feedback is fully addressed.

qayshp:
Superseding the earlier validation note: current head `0cd1cb84` contains the committed disable-failure case, and the focused suite re-passed **7/7**. The test confirms recovery stops at `enable → disable`, makes no second enable attempt, and preserves the exact existing disable error.

No test used a real disposable zrok token/environment, so this remains mocked command-boundary validation and is not an end-to-end live recovery claim.

PR #840, opened 2026-08-18
Fix development production build and migration startup

kevinwmerritt:
## Summary

- use `NodeJS.Timeout` for scheduled interval handles so the production TypeScript build accepts `clearInterval`
- give `AddExternalIdToContacts1750299580000` an explicit migration name so TypeORM can identify it after production minification

## Why

The packaged development build currently fails in two places:

1. TypeScript rejects `NodeJS.Timer` when it is passed to `clearInterval` with the current locked typings.
2. When upgrading an existing server database, the production-minified contact migration has the runtime class name `t`. TypeORM requires migration names to end in a 13-digit timestamp and aborts startup with:

   `t migration name is wrong. Migration class name should have a JavaScript timestamp appended.`

The older migrations already define an explicit `name` property for this reason.

## Verification

- ESLint passes for both changed files.
- `npm run build --workspace @bluebubbles/server` succeeds.
- A packaged ARM64 build starts successfully against an existing BlueBubbles 1.9.9 configuration database on macOS 15.7.7.

### DIFF
--- a/packages/server/src/server/databases/server/migrations/1750299580000-AddExternalIdToContacts.ts
+++ b/packages/server/src/server/databases/server/migrations/1750299580000-AddExternalIdToContacts.ts
@@ -1,6 +1,8 @@
 import { MigrationInterface, QueryRunner, TableColumn, TableIndex } from "typeorm";
 
 export class AddExternalIdToContacts1750299580000 implements MigrationInterface {
+    name = "AddExternalIdToContacts1750299580000";
+
     public async up(queryRunner: QueryRunner): Promise<void> {
         await queryRunner.addColumn(
             "contact",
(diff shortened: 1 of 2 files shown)


jattree:
Confirming this fixes startup of a packaged `development` build on **macOS 26.6.2 (Tahoe, 25G83), Apple Silicon (M2)**.

Without it I hit both problems described here. The TypeScript build failed on `clearInterval(this.handle)` in `ScheduledService.ts`. After working around that, the packaged app started once on a fresh install (no `config.db`, so `synchronize` is used). Every later launch hung at `Initializing server database...`. Running the binary with its console captured showed the swallowed rejection:

```
TypeORMError: t migration name is wrong. Migration class name should have a JavaScript timestamp appended.
    at MigrationExecutor.getMigrations (.../typeorm/migration/MigrationExecutor.js:415:55)
```

With the same two changes as this PR (`NodeJS.Timeout`, and `name = "AddExternalIdToContacts1750299580000"`), the build succeeds. The app then starts normally on repeated launches against the existing database, with HTTP service up and iMessage listeners running.

Build details: `development` @ e2a5bf6e plus #832, Node 20.11.1, electron-builder arm64 target.

PR #844, opened 2026-09-03
Make the iMessage DB watcher survive WAL replacement

### LINKED ISSUE #843: Private API kills Messages.app ~40ms after chat listeners arm, permanently killing the chat.db watcher
With `enable_private_api=1` and `private_api_mode=process-dylib`, every startup does this:

```
[16:10:14.283][info]  [BlueBubblesServer] Starting chat listeners...
[16:10:14.321][debug] [MessagesDylibPlugin] Killing process: Messages
```
```
[13:40:09.451][info]  [BlueBubblesServer] Starting chat listeners...
[13:40:09.484][debug] [MessagesDylibPlugin] Killing process: Messages
```

Killing Messages.app makes SQLite tear down `chat.db-wal`. When Messages relaunches the WAL is recreated with a **new inode**. `fs.watch()` on macOS binds to the inode, not the path, so ~40ms into every boot the watcher is watching a file that no longer exists.

`MultiFileWatcher` cannot recover:

- it discards every event that is not `change`, so the `rename` signalling the replacement is dropped
- its `error` handler only re-emits, never re-arms
- nothing in the receive path polls, so there is no fallback

**The server then receives nothing at all** until the process restarts, at which point it happens again. REST still works, so clients look healthy whenever they manually refresh, which disguises this as unreliable push.

### Evidence

20 days of debug logs, v1.9.9, ~174 process starts:

| signature | count |
|---|---|
| `New Message from` | **0** |
| `Message match found` | **0** |
| `Sending FCM notification (Priority: high)` | **0** |
| `Sending FCM notification (Priority: normal)` | 114 |

Incoming messages are the only `high` priority emit (`handleNewMessage`: `newMessage.isFromMe ? "normal" : "high"`), so zero of them means no inbound message was ever pushed in 20 days.

### Knock-on effect: attachment sends time out

`Server().messageManager.resolve()` is called inside `MessagePoller.poll()`. With the watcher dead the poller only runs its startup pass, so outgoing `MessagePromise`s are never resolved. `sendAttachmentSync` then waits its full 20 minute ceiling, which a reverse proxy cuts long before — Cloudflare returns 524 at 100s. Presents as "attachments fail but the server log is clean".

### Why this likely has not been reported

It requires the Private API, which requires disabling SIP. Without it Messages.app is never killed, the WAL is never recreated, and the fragile watch never actually breaks. On an affected machine it is 100% reproducible; otherwise invisible.

### Repro

1. SIP disabled, `enable_private_api=1`, `private_api_mode=process-dylib`
2. Start the server
3. Send the Mac a message from another device
4. `grep -c "New Message from" ~/Library/Logs/bluebubbles-server/main.log` -> `0`

### Suggested fix

Watch the containing directory rather than the individual files — directory inodes survive WAL replacement, which removes the whole failure class. Also re-arm on `error`, stop discarding `rename`, and have `IMessageListener.stop()` call `watcher.stop()` (it currently leaks watchers across service restarts).

PR to follow.

kevintran-git:
Fixes #843.

### The problem

`MultiFileWatcher` watches `chat.db` and `chat.db-wal` with `fs.watch()` on the file paths. On macOS that binds to the inode. SQLite deletes and recreates `chat.db-wal`, most reliably when Messages.app is killed — which the Private API does on every startup, roughly 40ms after the chat listener arms. Once that happens the watcher is permanently deaf and the server receives nothing until it restarts, whereupon it happens again.

It cannot self-heal: `rename` events are discarded, the `error` handler never re-arms, and nothing polls as a fallback.

### What it solves

Receiving survives the WAL being replaced, so the Private API no longer silently disables inbound messages. It also fixes the knock-on in #843 where outgoing `MessagePromise`s never resolve, because `messageManager.resolve()` only runs inside `MessagePoller.poll()`.

### How it is fixed

**`lib/MultiFileWatcher.ts`** — watch each file's parent directory instead of the file. Directory inodes are stable across WAL replacement, which removes the failure class rather than patching around it. Directory events are filtered to the watched files by name, then confirmed by comparing `mtimeMs`, `size` and `ino` before emitting, so the emitted `FileChangeEvent` shape is unchanged for consumers. Watches re-arm on error, bounded by a consecutive-failure count that resets on any successful event, so a permanently bad path cannot spin. `stop()` now fully tears down.

**`databases/imessage/listeners/IMessageListener.ts`**

- `stop()` now stops the watcher. It previously did not, leaking watchers across service restarts.
- the startup poll now emits its results. It previously ran with `emitResults = false`, which loaded every recent message into the dedupe cache without delivering any, so a restart silently swallowed the backlog instead of catching up.
- `getEarliestModifiedDate()` tolerates a missing file rather than throwing. `chat.db-wal` legitimately does not exist after a clean checkpoint.

### Not included

The startup ordering (listeners arming before dylib injection) is deliberately left alone. With a watcher that survives inode replacement it is no longer load-bearing, and sequencing service startup is a larger change worth discussing separately.

### Verification

On an affected machine `New Message from` and `Priority: high` were both 0 across 20 days and ~174 process starts. Both go non-zero on the first received message with this change.

### DIFF
--- a/packages/server/src/server/databases/imessage/listeners/IMessageListener.ts
+++ b/packages/server/src/server/databases/imessage/listeners/IMessageListener.ts
@@ -50,6 +50,7 @@ export class IMessageListener extends Loggable {
 
     stop() {
         this.stopped = true;
+        this.watcher?.stop();
         this.removeAllListeners();
     }
 
@@ -60,9 +61,13 @@ export class IMessageListener extends Loggable {
     getEarliestModifiedDate() {
         let earliest = new Date();
         for (const filePath of this.filePaths) {
-            const stat = fs.statSync(filePath);
-            if (stat.mtime < earliest) {
-                earliest = stat.mtime;
+            try {
+                const stat = fs.statSync(filePath);
+                if (stat.mtime < earliest) {
+                    earliest = stat.mtime;
+                }
+            } catch (error) {
+                this.log.debug(`Unable to stat ${filePath}: ${error}`);
             }
         }
 
@@ -70,13 +75,10 @@ export class IMessageListener extends Loggable {
     }
 
     async start() {
-        this.lastCheck = this.getEarliestModifiedDate().getTime() - 60000;
+        this.lastCheck = this.getEarliestModifiedDate().getTime();
         this.stopped = false;
 
-        // Perform an initial poll to kinda seed the cache.
-        // We'll use the earliest modified date of the files to determine the initial poll date.
-        // We'll also subtract 1 minute just to pre-load the cache with a little bit of data.
-        await this.poll(new Date(this.lastCheck), false);
+        await this.poll(new Date(this.lastCheck));
 
         this.watcher = new MultiFileWatcher(this.filePaths);
         this.watcher.on("change", async (event: FileChangeEvent) => {
--- a/packages/server/src/server/lib/MultiFileWatcher.ts
+++ b/packages/server/src/server/lib/MultiFileWatcher.ts
@@ -1,5 +1,6 @@
 import EventEmitter from "events";
 import fs from "fs";
+import path from "path";
 
 export type FileStat = fs.Stats | null | undefined;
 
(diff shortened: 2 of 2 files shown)


kevintran-git:
Retracting. I tested the mechanism instead of inferring it: the WAL inode never changes, `fs.watch` was never orphaned, and inbound messages push correctly. The real cause was a stray restart loop on my own machine. Sorry for the noise. — claude

zlshames:
I am actually in the process of rewriting the server from NodeJS -> Swift. And this was something that came up. The WAL may not exist when the server starts, so we need a way to detect it. I've already built out a system for that in the Swift version :)

https://github.com/BlueBubblesApp/bluebubbles-server/tree/claude/swift-server-migration-plan-wmoxm6

if you're curious

PR #848, opened 2026-09-15
fix: detect Live Photos before HEIC conversion

### LINKED ISSUE #847: bug: HEIC Live Photos not being detected
Some Live Photos (HEIC) are showing up as regular still images in the client app because the server is not correctly looking up the companion `.mov`

The server currently looks for the live photo companion `.mov` sibling **after** converting the image, which is written to a `Convert/` directory. The companion `.mov` actually lives next to the **original** source image, so the lookup misses and causes ``hasLivePhoto`` to be marked false.

Non-HEIC Live Photos, or requests that skip conversion, are not affected.

**Expected Behavior**

HEIC Live Photos show up as live photos in the client app

**Actual Behavior**

HEIC Live Photos are showing up as regular images

----

| iMessage on Mac with Live Photo icon | BlueBubbles with no Live Photo icon |
|---|---|
| Image | Image |

DeveloperBlue:
For HEIC Live Photos, the server currently checks for the companion ``.mov`` file after conversion, incorrectly inside the ``Convert/`` folder. This PR moves the check to before the conversion, so it correctly identifies the companion ``.mov`` next to the source HEIC image.

## Summary
- Compute `hasLivePhoto` before `convertImage` / `convertAudio` rewrites `filePath` into `Convert/`.
- Companion `.mov` detection looks next to the original Messages path, so post-conversion checks always missed HEIC Live Photos.
- Once `hasLivePhoto` is true, the client requests and plays the correct image and .mov pair on its own, no additional work is needed.

Fixes #847


DeveloperBlue:
I also have a two line client-side fix for clients that already have ``hasLivePhoto`` cached as false. Currently, resyncing on the app does not merge in the new attachment data; clients would have to clear their apps to see the live photo properly.

https://github.com/BlueBubblesApp/bluebubbles-app/compare/master...DeveloperBlue:bluebubbles-app:developerblue/live-photo-sync

Let me know if it's worth PRing

